PR #2, opened 2015-07-22
Increase the precision for #define float and doubles, and add an 'f' to the end of float string`

jkozdon:


dmed256:
Oops, missed the 'f' at the end

I changed it to use scientific + 8/14 instead of the 50/100 fixed digits in the last [night] push (just pushed it)

Thanks for point that bug out!

dmed256:
Here's the change:
  https://github.com/libocca/occa/blob/593bbfa97df97fe081fc6ec3f56c13e77dab6e44/include/occa/parser/tools.hpp#L36

dmed256:
@jkozdon, I just added it to [master] branch incase you needed it now

PR #8, opened 2015-08-25
Fix memory leaks

lcw:
Here are some changes to (hopefully) fix memory leaks in OpenCL, CUDA, and the C bindings.  I can use the address sanitizer in gcc to test the changes to OpenCL and the C bindings but not for CUDA.


dmed256:
Thanks!

PR #14, opened 2015-10-22
[Defines] Adding hypot function to the defines

jkozdon:
Adding hypot function to occa


dmed256:
Looks good, thanks!

jkozdon:
Perfect thanks!

jkozdon:
BTW, I like that you have eliminated the python script as it simplifies the inclusion of new functionality.

dmed256:
Thanks, it was pretty annoying

I would prefer to keep the source code in the .so but that didn't seem like
a good way to do it

I'll try to find a way to dump the source later, for now it's reading the
headers (also want to get rid of OCCA_DIR in the future)

PR #18, opened 2015-12-11
Add ispc addVectors example

lcw:
Hi @dmed256 and @tcew,

@inducer and I have put together an example of addVectors calling an ispc (https://ispc.github.io/) based kernel.  This is currently using the interface that @dmed256 has suggested.

Looking at it we both think that ispc might be a great addition to the occa backends to support CPU/Phi.  The kernel language is very close to CUDA/OpenCL and should be relatively straight forward to include in OCCA.  We think it might enable single source performant code in OCCA.

Best,
@lcw and @inducer

### DIFF
--- a/examples/addVectors/addVectors.ispc
+++ b/examples/addVectors/addVectors.ispc
@@ -1,0 +1,40 @@
+#define WGDIM0 16
+#define WGDIM1 1
+#define WGDIM2 1
+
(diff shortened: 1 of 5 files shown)


dmed256:
The example looks good, except I don't want to pull tasksys.cpp (similar to how we're not pulling OpenCL or CUDA headers)
I'm ok with putting a README saying there is a dependency on it though

PR #30, opened 2016-05-02
Detach methods for OCCA memory objects

pdhahn:
1. Added memory::detach() and memory::mappedDetach() methods that free the OCCA memory object without freeing the encapsulated device memory. This is a missing piece of OCCA interoperability for those cases when the user desires to recover and use the encapsulated device memory outside of OCCA without having to worry about scope or lifetime of the original OCCA memory object. Note the existing getMemoryHandle() method is used to access the encapsulated device memory before detaching.

2. Added some doxygen comments in header files. In particular, added description for memory::getMemoryHandle() to document what is actually returned for various OCCA devices, in order to facilitate the interoperability scenario.

3. Fixed problem with nonsense CPU frequency value displayed by occainfo on Linux. This includes a new cpu::getField() method that directly parses the output of a shell command (in this case, "cat /proc/cpuinfo") for a property (field) value, instead of relying on indirectly parsing it via invocation of scripts/shellTools.sh. The getField() method is valid for Linux, OSX, CygWin and MinGW, and is intended as a replacement for the Linux-pnly cpu::getFieldFrom() method that was broken and causing the problem with CPU frequency.

### DIFF
--- a/src/Serial.cpp
+++ b/src/Serial.cpp
@@ -86,25 +86,41 @@ namespace occa {
       while(fgets(buffer,bufferSize,fp)) {
 	// find line that matches "^[WS}*field[WS]*:"
 	str = buffer;
-	while(*str && isspace(*str)) ++str;
-	if (!*str) continue;
+	while(*str && isspace(*str)) {
+	  ++str;
+	}
+	if (!*str) {
+	  continue;
+	}
 	if (ignoreCase ?
 	    strncasecmp(str,field_str,field_len) :
-	    strncmp(str,field_str,field_len)) continue;
+	    strncmp(str,field_str,field_len)) {
+	  continue;
+	}
 	ptr = str + field_len;
-	while(*ptr && isspace(*ptr)) ++ptr;
-	if (*ptr != delimiter) continue;
+	while(*ptr && isspace(*ptr)) {
+	  ++ptr;
+	}
+	if (*ptr != delimiter) {
+	  continue;
+	}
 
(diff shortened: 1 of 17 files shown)


src/Serial.cpp
```diff
@@ -52,9 +61,80 @@ namespace occa {
...
+
+      int field_len = field.size();
+
+      const char *field_str = field.c_str();
+      char *ptr;
+      char *str;
+
+      std::string ret = "";
+
+      while(fgets(buffer,bufferSize,fp)) {
```
dmed256: Can you put loop statements in new lines (89-107)
For example

```C++
while(*ptr && isspace(*ptr)) ++ptr;
```

to

```C++
while(*ptr != '\0' && isspace(*ptr) {
  ++ptr;
}
```

dmed256:
Looks good, just have a comment about the code style and we're good to merge =)
Thanks!

src/Serial.cpp
pdhahn: Sure, no problem. I did not mean to "disrespect" your style. FYI I like 
structured style too. Just about the only /occasional/ exception being 
(for me) when I do a "one-liner" while() such as when scanning strings 
like this. And sometimes "if (...) continue; / break;" in loops.  Just a 
habit for me nowadays. But you are correct, I must remain consistent 
with your style when making any mods.

dmed256: No problem, thanks =)

PR #41, opened 2016-08-13
`occaDeviceBuildKernelFromString` with MPI bug

lcw:
There is a bug with at least the Serial and OpenMP modes that doesn't
allow for `buildKernelFromBinary` to be called from two processes on the
same kernel running at the same time (e.g., running the code with
`mpicc`).

These changes illustrate the bug and seem to fix it.

### DIFF
--- a/examples/addVectors/c/main.c
+++ b/examples/addVectors/c/main.c
@@ -43,6 +43,19 @@ int main(int argc, char **argv){
   addVectors = occaDeviceBuildKernel(device,
                                      "addVectors.okl", "addVectors",
                                      info);
+  const char* knl_str = " \n"
+    "kernel void addVectors(const int entries, \n"
+    "                       const float *a, \n"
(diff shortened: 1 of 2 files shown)


lcw:
This fix seems to work sometimes but not all of the time.  If I run the c addVectors example with 20 ranks a few times with

```
mpirun -np 20 ./main
mpirun -np 20 ./main
mpirun -np 20 ./main
mpirun -np 20 ./main
```

it crashes sometimes with

```
---[ Error ]--------------------------------------------
    File     : /Users/lucas/research/code/occa//src/tools.cpp
    Function : parseFileForFunction
    Line     : 826
    Error    : Could not find function [addVectors] in file [c691aac4102893c6/stringSource.okl]
```

dmed256:
I'll look into why its failing with MPI
My guess is I'm not locking the kernel properly

PR #42, opened 2016-11-22
Don't count mapped allocations

lcw:
Mapped allocations don't use device memory, so we will not count them
with the device memory counter.


lcw:
This seems to work for OpenCL but something is not right in the Serial backend.

dmed256:
Thanks :)

lcw:
No problem.  There still seems to be something wrong with the change for
a Serial device.  I haven't tracked that down yet.

PR #67, opened 2017-09-27
Detect new versions of XL compiler

jdahm:
This commit adds support for returning OCCA_IBM_VENDOR when compiling with newer versions of the XL compiler.


dmed256:
Thanks!

PR #68, opened 2017-10-11
[occa]: updating shelltools script for openCL on ROCm

nicholasmalaya:
This provides default paths for include and libraries for OpenCL in ROCm.

OCCA now successfully builds with, "make CXX=hcc" on my dual radeon (R9, Fiji Nano) box.
Most examples appear to be working correctly, and can run on either device. 

Note that this specifies the path to the OpenCL library as, /opt/rocm/opencl/lib/x86_64, not /opt/rocm/opencl/lib/, which did not pick up the library.


scripts/shellTools.sh
```diff
@@ -115,6 +116,7 @@ function defaultLibraryPath {
     mergedPaths+=":$OCCA_LIBRARY_PATH"
     mergedPaths+=":$LD_LIBRARY_PATH"
     mergedPaths+=":$DYLD_LIBRARY_PATH"
+    mergedPaths+=":/opt/rocm/opencl/lib/x86_64"
```
dmed256: Can you change x86_64 to *, that way it doesn't depend on the architecture

nicholasmalaya:
Done, also verified that the build is still working on my box.

nicholasmalaya:
Thanks!

PR #69, opened 2017-10-17
Silence warnings about multiplying bools

bobtwinkles:
Previously when compiling with -Wall vector.hpp would generate a large number of
warnings of the form:

```
occa/include/occa/defines/vector.hpp:1461:9: warning: ‘*’ in boolean context, suggest ‘&&’ instead [-Wint-in-bool-context]
   a.s15 *= b.s15;
```

This commit silences those warnings using GCC pragmas which will work with both
GCC and the mpic/mpic++ compilers.

### DIFF
--- a/include/occa/defines/vector.hpp
+++ b/include/occa/defines/vector.hpp
@@ -16,6 +16,8 @@
 namespace occa {
 #  endif
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wint-in-bool-context"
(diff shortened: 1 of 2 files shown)


dmed256:
Thanks for finding the bug and sending a PR!

The bool vector is [gone in 1.0](https://github.com/libocca/occa/blob/1.0/include/occa/vector.hpp), which is where all the updates are happening. Since the API changes a bit, let me know if it breaks any of your code for some help :)

PR #72, opened 2017-11-01
Fix the mode parameter in appendAvailableDevices

thilinarmtb:
We were getting an error when calling `occa::getDeviceList()`. It printed out an error saying
mode is not defined. So we made these changes. But still, `mode = OpenCL` and `mode = CUDA`
fails. `mode = Serial` works.

### DIFF
--- a/src/CUDA.cpp
+++ b/src/CUDA.cpp
@@ -1117,7 +1117,7 @@ namespace occa {
 
     for(int i = 0; i < deviceCount; ++i){
       device d;
-      d.setup("CUDA", i, 0);
+      d.setup("mode = CUDA", i, 0);
 
       dList.push_back(d);
     }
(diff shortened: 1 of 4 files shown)


dmed256:
oh, @tcew just pinged me about it and I added a hack fix
This is only an issue in 0.1 (master branch), I'll add the same method for 1.0 soon

dmed256:
Oops, the OpenCL and CUDA versions don't use 
```
setup("mode = OpenCL", p, d);
setup("mode = CUDA", d);
```
Since it uses a really oooooold way of initializing the devices

thilinarmtb:
I see. Thanks @dmed256.

dmed256:
@thilinarmtb: np, thanks for the PR! I saw the email after I pushed the change, sorry about that :(

thilinarmtb:
No problem at all. Thank you very much for the fix !!

PR #73, opened 2017-11-04
[WIP] Update C API for using in Nek5000/NekBone

thilinarmtb:
@dmed256 : I am planning to add a few functions to OCCA C API that we
need when porting Nek5000/Nekbone to be run with OCCA. We prefer
using C API in Nek5000/Nekbone rather than C++.

Are you fine with merging these changes into OCCA master?

I have added a function to get a list of available devices on the node so that
each MPI rank can select one. Let me know if this is okay. I will keep adding
some other functions we will be needing in future.


dmed256:
Looks good to me, thanks!

thilinarmtb:
Thanks for merging the PR David !!

PR #74, opened 2017-11-29
[BUILD] Add environment variables for CUDA paths

jkozdon:
We were having a problem on a system with multiple version of CUDA.


dmed256:
Hi @jkozdon 
Is there a reason not to use:
```
OCCA_INCLUDE_PATH
CPLUS_INCLUDE_PATH
C_INCLUDEPATH
INCLUDEPATH
```
for include paths and
```
OCCA_LIBRARY_PATH
LD_LIBRARY_PATH
DYLD_LIBRARY_PATH
```
for .so paths?

jkozdon:
You are correct. We were running with an old version of occa in our main code and the rearranging of the search path from f7f1e0dde2138270b57c7eb4dcf9d46830e6e19c do fix the problem using the `OCCA_INCLUDE_PATH` and `OCCA_LIBRARY_PATH`. I've just patched this change into our local occa version

dmed256:
Thanks for checking :)

PR #75, opened 2017-12-01
Add /opt/cuda to include and library search paths

bobtwinkles:
Arch Linux (and possibly other Linux distros) put all the CUDA files in `/opt/cuda`, so add those to the list.


dmed256:
Oops, can't merge collisions on the phone
I'll try later today

PR #81, opened 2018-01-28
Fix truncation problem in values shown for Memory when using occainfo.

pdhahn:
I installed a new graphics card with 2 GB memory but occainfo showed only 1 GB. The actual available memory was 1.95 GB which revealed that occainfo was truncating the fractional part. The culprit was the function stringifyBytes() in src/tools.cpp. I have modified that function so that it prints float values with 3 significant digits in each log decade.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <a name="OCCA"></a>
-## OCCA 
+## OCCA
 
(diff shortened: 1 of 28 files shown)


dmed256:
Ah, good catch!
Thanks for the PR :)

dmed256:
Added a569d1f to 1.0 to add the same fractional outputs

PR #100, opened 2018-03-31
initSignalHandling: do not attempt to catch SIGKILL

jedbrown:
SIGKILL cannot be caught so there is no point attempting to register
actions, which generates warnings in certain environments, such as
Valgrind:

  ==31674== Warning: ignored attempt to set SIGKILL handler in sigaction();
  ==31674==          the SIGKILL signal is uncatchable


dmed256:
Thanks!

PR #146, opened 2018-06-28
Makefile: fix shell diff

jedbrown:
### Description
Fixes rebuilding include/occa/defines/compiledDefines.hpp

### Checks
- [x ] Nothing got committed into `./lib` and `./obj`
- [ x] MIT License copyright in new files


dmed256:
Thanks!

PR #157, opened 2018-07-09
HIP Backend for OCCA

noelchalmers:
### Description

This pull adds AMD's HIP as a backend in OCCA's parser. In large part the HIP backend is identical to the CUDA backend as HIP was designed to be able to convert standard CUDA functions in place. The notable difference is in kernel::run() wherein HIP only supports passing arguments to kernels via the last parameter when launching modules, i.e. by wrapping a pointer to a void * struct 
 and packing with HIP_LAUNCH_PARAM_BUFFER_POINTER, HIP_LAUNCH_PARAM_BUFFER_SIZE, and HIP_LAUNCH_PARAM_END.

Also included is a fix to a __local variable scoping issue when using AMD's OpenCL compiler. By default OCCA will parse an okl such as 

```cpp
@kernel void foo() {
    for(int i=0;i<100;i++;@outer0) {
        @shared float s_a[10];
         for(int j=0;j<10;j++;@inner0) {
             s_a[j]+=1;
         }
    }
}
```

into OpenCL source 

```cpp
__global void foo() {
    {
        __local float s_a[10];
         {
             s_a[j]+=1;
         }
    }
}
```

However, AMD's OpenCL compiler with complain that __local variables must be declared in the outermost scope of a kernel function. In this pull, in the openclParser::setupKernels() function I've added a call to a new migrateLocalDecls function which moves all __local variable declarations to the top of the kernel function which fixes this error. 

### Checks


dmed256:
@noelchalmers Briefly looked through it and it looks pretty good!
I'll take a closer look once I get home, thanks again!

pdhahn:
I am concerned about the part that lifts _local variables. Are they actually being moved into host address space (inside the @kernel but outside the @outer nested kernel)?

noelchalmers:
@pdhahn I'm not sure I follow. In OpenCL mode, the __local variables are moved after kernels are split to directly after a statement like __global foo() {  so that they are in the outermost scope of the kernel. Can you clarify what you mean about the host address space?

pdhahn:
Also, I am using NVIdia GPUs with OCCA/OpenCL, so maybe your changes only apply to a problem that occurs when using AMD hardware?

noelchalmers:
@pdhahn Yes I believe that's correct. In my experience only the AMD OpenCL compiler complains about this type of scoping issue. It could likely be fixed on that end, but this is sufficient work around and doesn't alter any kernel behavior.

pdhahn:
Actually I corrected my comment to read migrate to "_global" instead of host address space. Is that what your code does?

noelchalmers:
No I don't believe so. In my example above, the parsed source file simply changes from this:

```cpp
__global void foo() {
    {
        __local float s_a[10];
         {
             s_a[j]+=1;
         }
    }
}
```

to this:

```cpp
__global void foo() {
     __local float s_a[10];
    {
         {
             s_a[j]+=1;
         }
    }
}
```

pdhahn:
Ah, got it. Thanks for the patient explanation..

noelchalmers:
No worries. My apologies if my initial explanation was unclear. It's a strange issue that really shouldn't exist, but lucky a fix is straightforward.

dmed256:
I put a few comments but overall it looks good, thanks a lot for adding this PR!

PR #165, opened 2018-07-19
XL compiler requires user to include cstdlib for system atof()

jdahm:
Small bug fix: XL compiler requires user to include cstdlib for system atof().


dmed256:
@jdahm Thanks!

PR #167, opened 2018-07-20
[Parser] Fixed register variables being migrated in OpenCL mode

noelchalmers:
### Description

Fixed an issue where my recent change to the OpenCL parser was mistakenly migrating statements. The old code searched for variable declarations with the "shared" in a kernel function and migrated them to the top of the kernel. However this would mistakenly migrate statements like 
`double r_q = s_q[i];`
when `s_q` was a shared array. This patch now checks if the variables being declared in the statement have the "shared" attribute themselves before migrating.

### Checks


dmed256:
Thanks!

PR #226, opened 2019-05-14
[HIP/OpenCL] Updates for HIP backend and updating deprecated OpenCL APIs

noelchalmers:
### Description

Updating the HIP backend to be in sync with CUDA backend support. Also added an OpenCL v2.0 check to avoid using deprecated `clCreateCommandQueue` (AMD's OpenCL complains about this). 

Side note: While testing the  @ restrict keyword in OpenCL mode, I noticed both Nvidia and AMD targets seem to want the restrict key word to be `__restrict__`, which means I had to change [line 21 of src/lang/modes/opencl.cpp](https://github.com/libocca/occa/blob/022b76829d43cbe20b719e6d5a54c9aff8fa178c/src/lang/modes/opencl.cpp#L21). I'm not sure if other platforms also need this keyword so I haven't included that change. 

### Checks
- [ ] MIT License copyright in new files

### DIFF
--- a/src/modes/hip/utils.cpp
+++ b/src/modes/hip/utils.cpp
@@ -16,7 +16,7 @@ namespace occa {
     bool init() {
       static bool isInitialized = false;
       if (!isInitialized) {
-        isInitialized = hipInit(0);
+        isInitialized = !hipInit(0);
       }
       return isInitialized;
     }
@@ -54,8 +54,8 @@ namespace occa {
 
     void enablePeerToPeer(hipCtx_t context) {
 
-      OCCA_HIP_ERROR("Enabling Peer-to-Peer",
-                     hipCtxEnablePeerAccess(context, 0) );
+      // OCCA_HIP_ERROR("Enabling Peer-to-Peer",
+      //                hipCtxEnablePeerAccess(context, 0) );
     }
 
     void checkPeerToPeer(hipDevice_t destDevice,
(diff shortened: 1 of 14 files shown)


src/modes/hip/utils.cpp
```diff
@@ -54,8 +54,8 @@ namespace occa {
 
     void enablePeerToPeer(hipCtx_t context) {
 
-      OCCA_HIP_ERROR("Enabling Peer-to-Peer",
-                     hipCtxEnablePeerAccess(context, 0) );
+      // OCCA_HIP_ERROR("Enabling Peer-to-Peer",
```
dmed256: Should we throw an exception here?

src/modes/hip/utils.cpp
```diff
@@ -151,31 +151,21 @@ namespace occa {
...
-
     occa::device wrapDevice(hipDevice_t device,
-                            hipCtx_t context,
                             const occa::properties &props) {
 
-      occa::properties allProps = props;
+      occa::properties allProps;
       allProps["mode"]     = "HIP";
-      allProps["device_id"] = -1;
+      allProps["deviceID"] = -1;
```
dmed256: I'm trying to stick with `snake_case` for property keys and `kebab-case` for command-like args.
```suggestion
      allProps["device_id"] = -1;
```

src/modes/hip/utils.cpp
noelchalmers: The hipCtx_t struct (analogue of a CUcontext) is actually marked as deprecated when compiling HIP for AMD targets. The reason is because its not really necessary for AMD's APIs. 

It is inconvenient to have this deprecate when compiling for Nvidia targets, though. But I assume it means the `hipCtxEnablePeerAccess` doesn't actually accomplish anything with AMD GPUs. AMD GPUs can access other GPU's memory when able (using flat addressing, I believe) by default. I can check on this though. 

In the meantime, I'm not sure if this should throw. Would it be better for switching between cuda<->hip if it just always returned successfully?

src/modes/hip/utils.cpp
noelchalmers: Good catch. I thought I saw this change in the cuda backend and I must have missed these.

dmed256:
Thanks for the fast updates!

PR #239, opened 2019-08-01
Copy to prefix instead of prefix/subfolder

thilinarmtb:
### Description

Copy bin, include, lib into PREFIX/ when installing.

### Checks


dmed256:
Thanks @thilinarmtb !

PR #240, opened 2019-08-02
soname changes for libocca.so

thilinarmtb:
### Description

* Added soname (install_name) field
* use .dylib as extension on OS/X

### DIFF
--- a/scripts/Makefile
+++ b/scripts/Makefile
@@ -171,6 +171,8 @@ pthreadFlag  = $(call compilerPthreadFlag)
 #---[ Flags and Libraries ]-----------------------
 OCCA_USING_VS := 0
 OCCA_UNSAFE   ?= 0
+soNameFlag=
+soExt=
 
 ifeq ($(usingLinux),1)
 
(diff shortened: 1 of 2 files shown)


scripts/Makefile
```diff
@@ -181,12 +183,16 @@ ifeq ($(usingLinux),1)
...
   linkerFlags += -lm -lrt -ldl
+  soNameFlag = -Wl,-soname,libocca.so
+  soExt = so
 
 else ifeq ($(usingMacOS),1)
 
 	OCCA_OS := OCCA_MACOS_OS
   flags += -Wno-deprecated-declarations
   linkerFlags += -framework accelerate -framework CoreServices
+  soNameFlag = -Wl,-install_name,@rpath/libocca.dylib
```
thilinarmtb: This could be
```sh
 -Wl,-install_name,$(PREFIX)/lib/libocca.dylib
```

as well. Not sure what to go for.

dmed256: We can merge this first and add `$(PREFIX)` in another commit if it becomes an issue

dmed256:
Thanks @thilinarmtb! 🎉

Issue #82, opened 2018-02-13
Allow memory allocation inside kernels


pdhahn:
FYI I can allocate OCCA device memory inside the occaKernel routine body (outside of actual kernel loops), which allows me to define temporary global device memory buffer(s) there instead of externally which then must be passed in as a parameter. This means I can keep the logic for allocation and management of temporary / work buffers used in kernels localized to the occaKernel routine code where it belongs.

I achieve this by passing in the _occaDevice_ pointer cast as a uint64_t, then calling a special-purpose externally-defined C++ allocation function ("_CreateOccaMemory_()") inside the routine body to return an _occaPointer_ to a buffer (global device memory). This allocation function simply takes the incoming _occaDevice_ pointer plus a number of elements argument, and calls its OCCA malloc() function. The return type is (for example) _occaPointer float*_. I also have a corresponding memory free function ("_ReleaseOccaMemory_()")

Anyway, it would be much better (and helpful for all OCCA programmers) if there were built-in support to do this without even having to explicitly pass in the _occaDevice_ pointer as a parameter from the outside. That pointer could be a hidden parameter generated and passed in by the OCCA build. And in terms of user syntax, the user could simply define global device memory via dynamically-sized arrays (auto-freed on return). Or similar to what I do, you could just have a new built-in OCCA allocator function (e.g., "_occaMalloc_()") that users call to create the global device memory, which is accessible in subsequent kernel loops. Of course, you would also want to provide an "_occaFree_()" function.

dmed256:
Do you mean something like this:
```cpp
@kernel void reverse(const int N,
                     int *vec) {

  int *temp = new int[N];

  for (int i = 0; i < N; ++i; @tile(16)) {
    temp[N - i - 1] = vec[i]
  }

  for (int i = 0; i < N; ++i; @tile(16)) {
    vec[i] = temp[i]
  }

  delete [] temp;
}
```
where `temp` is in reality allocated in the device?

pdhahn:
Sure that would be fine and true to C++. And not having to provide the "occaPointer" keyword explicitly ( i.e., "**occaPointer** int *temp = new int{N};") is nice, too. :-) In my suggestion, I just thought it might be easier for you to implement a new built-in allocator (and free) function, or maybe even dynamically-sized arrays. Although personally I do not like the latter because they are not in "real" C++ (not sure about more recent C++ versions).

dmed256:
Oh I mean the parser would translate
```cpp
int *temp = new int[N];
[...]

PR #260, opened 2019-11-30
[Core] Adds templated malloc

dmed256:


pdhahn:
Is this new malloc feature related to the API or to OKL kernel code (e.g., issue #82)?

dmed256:
@pdhahn New malloc feature which adds a templated malloc for easier use while keeping backwards compatibility

### Original malloc
```cpp
occa::memory mem = occa::malloc(10 * sizeof(float), src);
```

↓

### Initial dtype malloc
```cpp
occa::memory mem = occa::malloc(10, occa::dtype::float_, src);
```

↓

### New malloc
```cpp
occa::memory mem = occa::malloc<float>(10, src);
```

pdhahn:
Cool I like this. Thanks.

PR #266, opened 2019-12-04
[HIP] Adds support for hip-nvcc (Thanks @noelchalmers!)

dmed256:
### Description

Adds some fixes provided by @noelchalmers to help support HIP-NVCC

### DIFF
--- a/src/modes/hip/device.cpp
+++ b/src/modes/hip/device.cpp
@@ -57,7 +57,7 @@ namespace occa {
         compilerFlags = "-O3";
       }
 
-      kernelProps["compiler"]      = compiler;
+      kernelProps["compiler"]       = compiler;
       kernelProps["compiler_flags"] = compilerFlags;
 
       archMajorVersion = kernelProps.get<int>("arch/major", hipProps.major);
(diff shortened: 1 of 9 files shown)


src/modes/hip/device.cpp
```diff
@@ -225,12 +225,14 @@ namespace occa {
...
+        kernelProps.get<std::string>("hipcc_compiler_flags")
+      );
 
-      if (compiler_flags.find("-t gfx") == std::string::npos
-          && compiler_flags.find("-arch=sm") == std::string::npos) {
-        kernelProps["compiler_flags"] += ' ';
-        kernelProps["compiler_flags"] += kernelProps["compiler_flag_arch"];
+      if (hipccCompilerFlags.find("-t gfx") == std::string::npos
+          && hipccCompilerFlags.find("-arch=sm") == std::string::npos) {
+        kernelProps["hipcc_compiler_flags"] += ' ';
```
noelchalmers: I think this should be " " not ' '. The json complains about adding a single char type.

dmed256: Good catch!

PR #282, opened 2020-02-18
Issue#232: Assert chosen mode exists

suniljoshi82:
### Description
Throw exception using OCCA_ERROR if mode does not exist or not enabled


dmed256:
Thanks!

PR #283, opened 2020-02-19
Revert "Issue#232: Assert chosen mode exists"

dmed256:
Reverts libocca/occa#282


dmed256:
@jedbrown Reverting this change. We can leave it as optional since it's being used with defaults. Thanks for mentioning this!

jedbrown:
Thanks! Cc: @jeremylt

PR #289, opened 2020-03-04
[OpenMP] Avoid generating same kernel hash as Serial mode

noelchalmers:
### Description
Small change to inject information about using an OpenMP device when generating kernel hash. Without this, the same kernel hash is generated in both Serial and OpenMP modes and cached kernels are incorrectly found.

### DIFF
--- a/src/modes/openmp/device.cpp
+++ b/src/modes/openmp/device.cpp
@@ -12,6 +12,7 @@ namespace occa {
     hash_t device::kernelHash(const occa::properties &props) const {
       return (
         occa::hash(props["vendor"])
+        ^ occa::hash("openmp")
         ^ props["compiler"]
         ^ props["compiler_flags"]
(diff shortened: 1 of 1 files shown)


dmed256:
Good catch! This was in the device at some point and I guess it got removed:

```cpp
  void device::setupKernelInfo(const occa::properties &props,
                               const hash_t &sourceHash,
                               occa::properties &kernelProps,
                               hash_t &kernelHash) const {
    assertInitialized();

    kernelProps = kernelProperties() + props;
    kernelProps["mode"] = mode();

    kernelHash = (
      hash()
      ^ modeDevice->kernelHash(kernelProps)
      ^ kernelHeaderHash(kernelProps)
      ^ sourceHash
    );

    kernelHash = applyDependencyHash(kernelHash);
  }
```

Thanks!

Issue #297, opened 2020-03-17
Unsupported flags passed to Intel compilers with CMake


SFrijters:
When using CMake to build occa with a somewhat recent version the Intel compiler (19.0.0.20181018) I get a large number of warnings `icpc: command line warning #10148: option '-Wno-c++11-long-long' not supported`.

Maybe it's better to avoid these flags if USE_INTEL is set to ON.

PR #298, opened 2020-03-17
[CMake] Do not pass incompatible flags to Intel compilers

SFrijters:
### Description

Attempts to solve #297


dmed256:
Awesome, thanks for finding this issue + the fix!

PR #299, opened 2020-03-18
Cmake fixes

amikstcyr:
### Description
A couple of annoyances with recent cmake version. 
This also fixes a complaint from Intel compiler.


dmed256:
Thanks!

PR #300, opened 2020-03-23
ADD: c language support for serial mode

amikstcyr:
### Description

Adds basic support for c files for serial mode when the OKL parser is not used.


dmed256:
Closing because of #306

PR #301, opened 2020-03-23
Const struct

amikstcyr:
### Description
okay this is the last PR + const for occaStruct. Previous PR on adding c compiler can be ignored.


dmed256:
Closing because of #306

PR #310, opened 2020-04-07
Add CTest support

### LINKED ISSUE #309: Add tests to CTest
With the recent CMake support it would be nice to also be able to run the test suite using CTest.

SFrijters:
### Description

Attempts to close #309 .

This also includes some general improvements to the CMake files, and some fixes that were exposed during the implementation and testing of the CTest.

One bugfix that may need discussion in particular is db7e8e4.

Please let me know if you want me to split things up differently.

EDIT: updated commit hash after rebase


examples/cpp/13_native_cuda_kernels/main.cpp
```diff
@@ -31,7 +31,6 @@ int main(int argc, const char **argv) {
   occa::memory o_ab = device.malloc<float>(entries);
 
   // Compile a regular CUDA kernel at run-time
-  occa::properties kernelProps;
```
dmed256, line 34: Oops, the `occa::properties` above should be `deviceProps` and not `kernelProps`

include/occa/scripts/tests/compiler.cpp
```diff
@@ -19,6 +19,10 @@ int main(int argc, char **argv) {
 #elif defined(__ICC) || defined(__INTEL_COMPILER)
   return OCCA_INTEL_VENDOR;
 
+// Clang also defines __GNUC__, so check for __clang__ first
```
dmed256, line 22: This was set at the bottom since other compilers are wrappers on `clang`. Maybe we should move the `GNU` checks even lower

src/io/utils.cpp
```diff
@@ -167,7 +177,7 @@ namespace occa {
       expFilename = fo.expand(expFilename);
 
       if (makeAbsolute && !isAbsolutePath(expFilename)) {
-        expFilename = env::PWD + expFilename;
```
dmed256, line 170: Can we initialize `env::PWD` to be the result of `currentWorkingDirectory` rather than calling this function every time?

SFrijters:
Thanks for the quick review. I'll get on the requested changes today.

examples/cpp/13_native_cuda_kernels/main.cpp
SFrijters: Ah, yes. I was blindly following compiler warnings, but I have the proper fix now.

include/occa/scripts/tests/compiler.cpp
SFrijters: Tricky... moved the clang check to its original place and GNU below that.

src/io/utils.cpp
SFrijters: Ok, did that now. I've also taken the liberty of renaming env::PWD to env::CWD to avoid possible confusion since it now doesn't reflect the actual ${PWD} variable anymore.

SFrijters:
I've addressed your concerns I hope and merged the changes into the relevant commits. Currently running tests locally before updating this PR.

Could the Travis pipeline be expanded in the future to also test the CMake/CTest path?

SFrijters:
Updated my branch; works on my end :)

dmed256:
@SFrijters Awesome, thanks for the quick fixes! I'll try running the tests tonight and merge if they pass. Not sure why Travis didn't run them for this PR :(

PR #318, opened 2020-04-14
[C] Use bool from stdbool.h for boolean types

awehrfritz:
Use `bool` from `stdbool.h` for boolean types. 

This header was introduced in C99, for more details see:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdbool.h.html

I tested this change by running the C examples, which use this functionality. All work as expected.


dmed256:
Awesome, looks so much better now :)

PR #321, opened 2020-04-14
[CMake] Let CMake determine which compiler flags should / can be used

### LINKED ISSUE #320: [CMake] Incompatible flags passed to compiler
The recent commit 7b2e817 introduced a compiler flag `-Wno-c++11-extensions` that is not accepted by at least the Intel compiler.

I have a PR coming soon that generalizes the checks for possible compiler flags and uses some more CMake standardization.

SFrijters:
## Description

Attempts to fix #320:

* Set the C++11 standard globally.
* Same for no C++11 extensions.
* Add functions to test if flags are accepted by the compiler and add only those that are.
* Add more warning flags.

Although I have automated tests set up for clang / gnu / Intel I only use Linux, so maybe @JamesEggleton can check if this works for him on MacOS?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,12 @@ project (occa
   HOMEPAGE_URL "https://github.com/libocca/occa/"
   LANGUAGES C CXX)
 
+set(CMAKE_CXX_STANDARD 11)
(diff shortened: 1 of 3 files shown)


dmed256:
Cool, thank you for the Cmake updates!

PR #323, opened 2020-04-15
[Cmake] HIP/CUDA/OpenCL on AMD and NV support through CMake

noelchalmers:
## Description

This PR adds HIP and OpenCL builds on AMD and NV machines, as well as fixes some CUDA build issues when cuda libraries are installed in different locations. Tested:

HIP/OpenCL with ROCm 3.3 on MI-60
HIP/CUDA/OpenCL with CUDA 9.0 in /usr/local/cuda on TitanV
CUDA/OpenCL with CUDA 9.1 installed through apt on GFX 1070.


dmed256:
:fire: Awesome, thank you for the CMake update!!

PR #328, opened 2020-04-20
[Core] Tweak behaviour of isSafeToRmrf

### LINKED ISSUE #326: testRmrf fails
This is just a small thing, but for my CI (jenkins) user `testRmrf` fails.
I've looked into the implementation and my environment and it happens because the `jenkins` user is a local user and has a `$HOME` of `/foo/home/jenkins` instead of `/home/jenkins`, which I think is the tacit assumption in the `isSafeToRmrf` function.

In general, it feels to me like that function is giving a false sense of security, or could at least use some additional documentation, since from the function name I would never have guessed the criteria.

Also, in some sense it may also be too strict, because something like `/tmp/occa` should be 'safe' to remove even though it only has depth 2. Maybe it would be nicer to check if the normalized path is below the user's actual home directory instead, or something like that? That still wouldn't cover the `/tmp` case though.

SFrijters:
A path is now considered safe to remove if any component of
it is 'occa' or '.occa', and if the final depth > 1.

## Description

Attempts to fix #326 .

### DIFF
--- a/src/tools/sys.cpp
+++ b/src/tools/sys.cpp
@@ -272,9 +272,13 @@ namespace occa {
       int depth = 0;
 
       strVector path = split(expFilename, '/', '\0');
+      bool foundOcca = false;
       const int pathSize = (int) path.size();
       for (int i = 0; i < pathSize; ++i) {
         const std::string &dir = path[i];
+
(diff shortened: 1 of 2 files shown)


dmed256:
Looks good, thanks!

PR #329, opened 2020-04-20
Fix compiler warnings

### LINKED ISSUE #327: Compiler warnings with new CMake builds
With the recently added warning flags in the CMake build, various compilers are complaining about things.

I'm working on a PR to fix these issues.

SFrijters:
## Description

Attempts to fix #327 .

I've tried to restructure my changes into separate commits to address separate issues for easier reviewing My commit c57a0bd in particular is rather opinionated, so that one probably warrants extra attention.

In general I'm trying to make the compilers as happy as possible by making actual changes, but some warnings have been suppressed explicitly.

This has been tested with Intel 2019, GCC 8.3 and Clang 9.


noelchalmers:
For some reason with GCC 5.4 the fix to suppress the OpenCL polyfill unused variables doesn't seem to work. Still spits a bunch of warnings.

SFrijters:
Hm, looks like that might be a compiler bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69967
Fixed in 6.x, but not in 5.4.

SFrijters:
Rebased on new master and changed new function `isBitwiseEqual` to `areBitwiseEqual` for more consistency with the existing family of `areEqual` functions.

dmed256:
@SFrijters Cool thanks, I'll take a look at the code changes tonight!

src/lang/primitive.cpp
```diff
@@ -248,8 +249,8 @@ namespace occa {
     case primitiveType::uint32_ : return primitive(!p.value.uint32_);
     case primitiveType::int64_  : return primitive(!p.value.int64_);
     case primitiveType::uint64_ : return primitive(!p.value.uint64_);
-    case primitiveType::float_  : return primitive(!p.value.float_);
-    case primitiveType::double_ : return primitive(!p.value.double_);
+    case primitiveType::float_  : OCCA_FORCE_ERROR("Cannot apply operator ! to float type");   break;
```
dmed256, line 252: These might be potentially used int he preprocessor, so we can't take them away

SFrijters: Ok, I can do something like `return primitive(areBitwiseEqual(p.value.float_, 0))`?
Is it just this one, or do you want similar changes for `&&` and `||`? Since `&` and `|` are already disallowed in the current implementation I would say no?

dmed256: Maybe for these operations, we can first cast the `float`/`double` as `int`

pdhahn: Ha ha, for float values, what is "zero" anyway?

My preference is the "...cannot apply operator ! to float type". Just leave it up to the user to explore the possibilities when a float type is involved based on his program logic. The user should already know he is in trouble if he's thinking about using logical not with a float value!

Yet I can't help but raise the question, for this primitive (`not`), is a sort of "near-`!`" acceptable, which checks for zero within _macheps_? e.g.,

`case primitiveType::float_  : return primitive( ((float)1.0 + p.value.float_) == (float)1.0 )`

My view: this is **not** acceptable for general use cases at this level. But others may disagree.

FYI possibly another approach to `!` based on an equality to zero comparison without employing the explicit `==` (i.e., _partial-order_ trick):

`case primitiveType::float_  : return primitive( !(p.value.float_ < -p.value.float_ || -p.value.float_ < p.value.float_) )`

All the above _mutatis mutandis_ for the double case.

src/c/types.cpp
```diff
@@ -646,7 +651,7 @@ const int OCCA_PROPERTIES    = occa::c::typeType::properties;
 const occaType occaUndefined  = occa::c::undefinedOccaType();
 const occaType occaDefault    = occa::c::defaultOccaType();
 const occaType occaNull       = occa::c::nullOccaType();
-const occaUDim_t occaAllBytes = -1;
+const occaUDim_t occaAllBytes = static_cast<occaUDim_t>(-1);
```
awehrfritz, line 654: Is this really necessary? I would hope the compiler can figure this out on its own, at least for such primitive types.

SFrijters: Clang reports it as an implicit conversion, so yes, it's needed to fix the warning.

awehrfritz: OK. I just think this makes the code look ugly and cluttered. IMHO, code should be easy to read and intuitive, this is far from that. I would rather consider to add an exception (i.e. suppress the warning) for Clang than adding something like this to the code. Just my two cents...

dmed256: How about we define a const value once and reuse it?
```cpp
const occa::udim_t UDIM_DEFAULT = static_cast<occa::udim_t>(-1);
```

This would also make the code a bit clearer in places where `-1` is set as the default because of how high the number is

src/lang/primitive.cpp
dmed256: The argument made here is that we want to support as much C/C++ as possible, regardless of whether it's considered "good practice" or "bad practice". I 100% agree we shouldn't be doing exact comparisons with `float` or `double` "in practice". However, the goal here is language coverage.

---

TIL: Looks like floats aren't supported in the preprocessor!

```
main.cpp:4:6: error: floating constant in preprocessor expression
 #if !0.0
```

src/c/types.cpp
awehrfritz: @dmed256 that's probably a sensible compromise. I like the simple `-1` since it's rather intuitive 
 and easy to understand as is, but having a predefined constant works as well - anything that avoids cluttering the source code with type casts.

src/lang/primitive.cpp
SFrijters: Since the preprocessor won't accept them anyway, can we keep these operators unimplemented for float/double? That would still be my preferred solution.

src/c/types.cpp
SFrijters: I've redone the relevant commit to use the const UDIM_DEFAULT, as suggested.

src/lang/primitive.cpp
dmed256: Sounds good!

dmed256:
Thank you to everyone that made this possible!

PR #334, opened 2020-04-27
[Build] Add -march=native to GCC/LLVM flags

awehrfritz:
## Description

This adds the `-march=native` option to the GCC/LLVM flags, similar to the options for INTEL, PGI, PATHSCALE, and others.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,12 +16,14 @@ jobs:
             CC: gcc-8
             CXX: g++-8
             OCCA_COVERAGE: 1
+            OCCA_CI_FIX_GCC_FLAGS: 1
 
           - name: (Ubuntu) gcc-9
             os: ubuntu-18.04
             CC: gcc-9
             CXX: g++-9
             OCCA_COVERAGE: 1
+            OCCA_CI_FIX_GCC_FLAGS: 1
 
           - name: (Ubuntu) clang-6
             os: ubuntu-18.04
@@ -54,6 +56,7 @@ jobs:
       CC: ${{ matrix.CC }}
       CXX: ${{ matrix.CXX }}
       OCCA_COVERAGE: ${{ matrix.OCCA_COVERAGE }}
+      OCCA_CI_FIX_GCC_FLAGS: ${{ matrix.OCCA_CI_FIX_GCC_FLAGS }}
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,6 +61,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: Compiler info
+      run: make -j info
+
     - name: Compile library
       run: make -j
 
--- a/Makefile
+++ b/Makefile
@@ -195,6 +195,7 @@ info:
 	$(info LDFLAGS        = $(or $(LDFLAGS),(empty)))
 	$(info --------------------------------)
 	$(info compiler       = $(value compiler))
+	$(info vendor         = $(vendor))
 	$(info compilerFlags  = $(compilerFlags))
 	$(info flags          = $(flags))
 	$(info paths          = $(paths))
--- a/include/occa/scripts/shellTools.sh
+++ b/include/occa/scripts/shellTools.sh
@@ -381,13 +381,13 @@ function compilerReleaseFlags {
     local vendor=$(compilerVendor "$1")
 
     case "$vendor" in
-        GCC|LLVM)   echo " -O3 -D __extern_always_inline=inline"     ;;
-        INTEL)      echo " -O3 -xHost"                               ;;
-        CRAY)       echo " -O3 -h intrinsics -fast"                  ;;
-        IBM)        echo " -O3 -qhot=simd"                           ;;
-        PGI)        echo " -O3 -fast -Mipa=fast,inline -Msmartalloc" ;;
-        PATHSCALE)  echo " -O3 -march=auto"                          ;;
-        HP)         echo " +O3"                                      ;;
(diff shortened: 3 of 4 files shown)


awehrfritz:
I usually would also specify `-mtune=native` for GCC, but I could not find that option for LLVM (llc). We could split this up though to add it for GCC.

There is an interesting article that gives a bit more background infos on this: https://lemire.me/blog/2018/07/25/it-is-more-complicated-than-i-thought-mtune-march-in-gcc/

awehrfritz:
I am not entirely sure why the tests with GCC 8/9 on Ubuntu fail, but I kinda suspect it might have to do with the compiler not being able to detect the correct CPU flags when using `-march=native`.

The tests with GCC on MacOS worked. But the CI runners are different for MacOS and Linux, the former is hosted by MacStadium and the latter on the Azure cloud: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners

awehrfritz:
I was able to get the tests to pass, but had to resort to a not so nice workaround: I had to add an environment variable to the GitHub Actions setup which I then used to filter-out the `-march=native` option. Maybe someone else has a better way of doing this...

While debugging this I printed out what the compiler actually detects when using this option on the CI runner. It appears that for Ubuntu with gcc-8, the compiler identifies the architecture as `knl` and with gcc-9 as `broadwell`. This is how to print out the target architecture with gcc/g++:
``` sh
g++ -march=native -Q --help=target | egrep -- '-march=|-mtune'
```

This exercise also seems to reveal that the MacOS test with gcc in fact uses clang in the background after all, here the error message I get on the MacOS runner when using the above options:
``` sh
2020-05-03T09:21:39.8239770Z env:
2020-05-03T09:21:39.8239910Z   CC: gcc
2020-05-03T09:21:39.8240030Z   CXX: g++
2020-05-03T09:21:39.8240120Z   OCCA_COVERAGE: 1
2020-05-03T09:21:39.8240240Z ##[endgroup]
2020-05-03T09:21:46.6330430Z clang: error: unsupported option '--help=target'
2020-05-03T09:21:46.6405790Z clang: error: no input files
2020-05-03T09:21:46.6499080Z 'GCC w/o -march=native: ' 
2020-05-03T09:21:47.5515180Z clang: error: unsupported option '--help=target'
2020-05-03T09:21:47.5519260Z clang: error: no input files
2020-05-03T09:21:47.5522250Z 'GCC w/  -march=native: ' 
```
Maybe we should add an info statement to the make file which prints out the compiler vendor as identified by compiling the `compiler.cpp` test... probably more reliable than whatever information is passed along from the CI runner.

awehrfritz:
I just added the print statement for the compiler vendor and this shows that the run on MacOS with gcc in fact uses Clang/LLVM and not GCC.

At the moment, the two tests with GCC both fail with the `-march-native` option (probably due to the compiler not being able to detect the architecture properly). Thus, we could filter-out the option for all GitHub CI runs using GCC and remove the extra env variable again.

pdhahn:
@awehrfritz - Is gcc perhaps soft-linked to clang on your MacOS system, and if so, is that "normal"? Or perhaps is gcc missing and cmake machinations discover clang and use it instead? I am just trying to ask if it is worth a better understanding of what is going on here, before wiring in more changes. Caveat: I am not a MacOS man so I really don't know the platform. Sorry if my jumping in here is ignorant but just throwing in my 2 cents. Your answers if nothing else will serve to edify.

awehrfritz-unsw-edu-au:
@pdhahn This is all related to the GitHub CI runners, not my own system. You can have a look at the logs yourself - I couldn't get much out of them though.

> Is gcc perhaps soft-linked to clang on your MacOS system, and if so, is that "normal"?

I have no idea why or how this is done on MacOS, I don't use a Mac for software development nor do I have access to a Mac to test this myself. I just stumbled across this when debugging the Linux issue. However, even if I had a Mac, that wouldn't help much with the CI runners as the configuration might be quite different there. I believe David would know more about that as he may have admin access to the CI runners, and he has a Mac to test it locally. According to the [docs](https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md) GCC should be installed on the MacOS instances.


> Or perhaps is gcc missing and cmake machinations discover clang and use it instead? 

This has nothing to do with cmake, this is with good old `make`.


> I am just trying to ask if it is worth a better understanding of what is going on here, before wiring in more changes. 

Well, the additional print statement I added, identified the problem quite well. And I think printing the compiler vendor is useful in any case.

awehrfritz:
@pdhahn This is all related to the GitHub CI runners, not my own system. You can have a look at the logs yourself - I couldn't get much out of them though.

> Is gcc perhaps soft-linked to clang on your MacOS system, and if so, is that "normal"?

I have no idea why or how this is done on MacOS, I don't use a Mac for software development nor do I have access to a Mac to test this myself. I just stumbled across this when debugging the Linux issue. However, even if I had a Mac, that wouldn't help much with the CI runners as the configuration might be quite different there. I believe David would know more about that as he may have admin access to the CI runners, and he has a Mac to test it locally. According to the [docs](https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md) GCC should be installed on the MacOS instances.


> Or perhaps is gcc missing and cmake machinations discover clang and use it instead? 

This has nothing to do with cmake, this is with good old `make`.


> I am just trying to ask if it is worth a better understanding of what is going on here, before wiring in more changes. 

Well, the additional print statement I added, identified the problem quite well. And I think printing the compiler vendor is useful in any case. I think as a next step, someone with a Mac would need to figure out the issue with the MacOS runner and GCC. Depending on whether we keep the MacOS gcc runner or not, we can then tweak the workaround.

awehrfritz:
Superseded by PR #340

Issue #104, opened 2018-04-26
Suggestions for Fortran API


dmed256:
It would be nice to revive the Fortran API found in the (libocca/occa-fortran)[https://github.com/libocca/occa-fortran] repo

Unfortunately I'm a Fortran noob, so I'm looking for suggestions on how to make the Fortran API look nice and mirror the C API (or cleaner if we can!). Questions include:

- Should we aim for F77 or a newer version?
- How does `NULL` work in Fortran? (we use it for some defaults)
- How do we wrap the `occaInt`, `occaDouble`, ... C objects?

nicholasmalaya:
* I'm a huge fan of SWIG, and could help provide an example. This is an easy way to generate interfaces to a vast multitude of languages, more than 'just' fortran. Would introduce a dependency into OCCA but it could be made command line optional. 
* F2003 is probably best, but F90 should be considered 'minimal'. F77 is prehistoric.
* F2003 does have a concept of a NULL pointer, but I don't know if that can be passed, C-like

Just my two cents. Great idea!

-NM

dmed256:
I would prefer we tailor the APIs based on the language, but I can take a look at SWIG to start with!

If no one else complains about F2003 being too 'new', I'll aim for that once I (or someone else) starts the Fortain API

pdhahn:
I suppose my minimum recommendation would be f95. This keeps the parser challenges manageable while accommodating what I believe are the interests of most Fortran programmers still actively developing or maintaining(!) code in various application domains where Fortran was/is the language of choice. Writing a parser for f2003 might be more time-consuming and not worth the extra effort if the vast majority of legacy Fortran code still being used, maintained and suitable for _OCCA-izing_ does not require it.

On the other hand, if OCCA can parse f2003 /f2008, then that would be a very good reason for coders to (re)write code using f2003/f2008 features moving forward. :-)

As for SWIG, per their site page: "SWIG is typically used to parse C/C++ interfaces and generate the 'glue code' required for the above target languages to call into the C/C++ code." I did not see Fortran in the list of languages(?).

dmed256:
@pdhahn This is for the API, not for a Fortran Kernel Language (like OFL in the past)

pdhahn:
Oops, sorry for missing the point (yes I was thinking a revival of "OFL" was in the offing for 1.0)!

dmed256:
np, thought that counts :)

dmed256:
[...]

PR #341, opened 2020-05-10
Add Fortran interface

awehrfritz:
## Description

This add a Fortran interface for OCCA and respective tests and examples. 

The implementation utilises the Fortran 2003 `ISO_C_BINDING` module. Thus a [Fortran 2003 compliant compiler](http://fortranwiki.org/fortran/show/Fortran+2003+status) is required. The interface relies on OCCA's C interface and is sthus limited to the functionality exposed via the C interface. The Fortran implementation is in most cases a simple `interface` to the C functions, thus simply providing a standardised way for the compiler to call the C functions from a  Fortran program.

The Fortran interface is compiled into a shared library `libocca_fortran.so`, which can be simply linked to an application intending to use OCCA. Alternatively, the Fortran modules can be directly included in a statically compiled program, as demonstrated in `examples/fortran/03_static_compilation`.

This resolves issue #104.

### Future work
While this initial implementation is fully functional, a subsequent revision could include additional wrapper/convenience functions to make the interface more "Fortran-like". 
The following lists a few examples in this regard:
1. The Fortran-to-C string conversion could be hidden by implementing a small shim layer around the `interface` in order to hide the `F_C_str` function in:
``` Fortran
device = occaCreateDeviceFromString(F_C_str(info))
```
2. Another example would be the Fortran pointer assignment for UVA memory, which could be wrapped in a single function call. Currently this requires the following:
``` Fortran
  type(C_void_ptr)       :: a 
  real(C_float), pointer :: a_ptr(:)

  a  = occaTypedUMalloc(entries, occaDtypeFloat, C_NULL_ptr, occaDefault)

  if (C_associated(a)) then
    call C_F_pointer(a,a_ptr,[entries])
  end if
```
3. The call to `C_loc` in 
``` Fortran
call occaCopyPtrToMem(o_b, C_loc(b), occaAllBytes   , 0_occaUDim_t, occaDefault)
```
could be hidden in a convenience function/wrapper analog to the proposed shim layer in 1.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,6 +60,7 @@ jobs:
       FC: ${{ matrix.FC }}
       OCCA_COVERAGE: ${{ matrix.OCCA_COVERAGE }}
       OCCA_FORTRAN_ENABLED: ${{ matrix.OCCA_FORTRAN_ENABLED }}
+      FORTRAN_EXAMPLES: ${{ matrix.OCCA_FORTRAN_ENABLED }}
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,6 +57,7 @@ jobs:
     env:
       CC: ${{ matrix.CC }}
       CXX: ${{ matrix.CXX }}
+      FC: ${{ matrix.FC }}
       OCCA_COVERAGE: ${{ matrix.OCCA_COVERAGE }}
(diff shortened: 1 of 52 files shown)


dmed256:
:fire: :fire: :fire: This is pretty amazing, I'll take some time this week to check it out. Thank you!!

dmed256:
Sorry it took a while. I have a few changes I would like to make but nothing blocking.

Thanks again for all your hard work!

PR #342, opened 2020-05-22
Enable use of CMake in github actions .yml

gionellef:
## Description

Creates a "build" directory outside the project (not in source)
Uses existing CMakeLists.txt, which includes CTest driver, for compilation and testing.


.github/workflows/build.yml
```diff
@@ -18,6 +18,15 @@ jobs:
...
             OCCA_COVERAGE: 1
             OCCA_FORTRAN_ENABLED: 1
+            useCMake: false
+
+          - name: (Ubuntu) CMake + gcc-8
+            os: ubuntu-18.04
+            CC: gcc-8
+            CXX: g++-8
+            OCCA_COVERAGE: 1
+            OCCA_FORTRAN_ENABLED: 1
```
awehrfritz, line 28: The Fortran parts don't include CMake files, neither the library nor the examples/tests. I suggest to disable the Fortran interface in the CMake runs for the time being.

.github/workflows/build.yml
```diff
@@ -18,6 +18,15 @@ jobs:
             FC: gfortran-8
             OCCA_COVERAGE: 1
             OCCA_FORTRAN_ENABLED: 1
+            useCMake: false
```
awehrfritz, line 21: This line is unnecessary and can be removed.

.github/workflows/build.yml
```diff
@@ -81,4 +107,4 @@ jobs:
       run: ./tests/run_examples
 
     - name: Upload code coverage
-      uses: codecov/codecov-action@v1
+      uses: codecov/codecov-action@v1
```
awehrfritz, line 110: There needs to be an empty line at the end of the file.

.github/workflows/build.yml
```diff
@@ -70,9 +79,26 @@ jobs:
 
     - name: Compile library
       run: make -j
+      if: ${{ !matrix.useCMake }}
```
awehrfritz, line 82: Please add this line also to the `Compiler info` step. Printing the make settings for a CMake build doesn't make much sense.

dmed256:
@gionellef Thanks for making all of these changes! Looks like the failures are from warnings that are fixed in #329. I think we can merge this soon though, excited about having CMake as part of the test suite :)

dmed256:
@gionellef Looks like the tests are running and passing! However, there seems to be an issue when running the example codes from the CMake build directory

gionellef:
Oh wow I didn’t realize the problem could be that simple :D it was my pleasure, thanks a lot for letting me learn!!

dmed256:
@gionellef will you have time to finish this PR?

PR #371, opened 2020-06-15
[CMake] Place compiledDefines header in Cmake build directory, rather than source tree. 

noelchalmers:
## Description

This PR is to move the install location of the `compiledDefines.hpp` header to the cmake build folder, rather than placing it in the OCCA_DIR/include folder. This behavior makes the cmake build system align better with the usual cmake workflow.


dmed256:
@noelchalmers Oh interesting, so this only affects the CMake build?

noelchalmers:
Ya this only affects the cmake build. The normal build system just has a repeated include path on the JIT compile lines. In theory, we could back port the paths so that compiledDefines.hpp could be moved and the builds would still work, but seeing as how it's manually placed in occa/include I don't think that's necessary.

Issue #373, opened 2020-06-16
Exceptions and extern "C"


SFrijters:
It seems throwing exceptions from functions marked 'extern "C"' is undefined behaviour.

This causes some tests that use ASSERT_THROW to fail when compiled with the PGI (20.1-0) compiler, since the exception is not propagated/caught properly anymore and std::terminate is called. For testing purposes I looked at the c-device test and moved `occaCreateDevice` out of the 'extern "C"' part of the code and then the test worked correctly.

Any thoughts on a proper way to fix this? It seems kind of involved to do it nicely.

Issue #372, opened 2020-06-16
Fortran examples segfault with Intel compiler


SFrijters:
The Fortran examples are crashing for me in the occaKernelRun function when using the (19.0.5.281) Intel compiler. It is fine with gcc+gfortran and clang+gfortran.

As far as I could find out, calling variadic C functions from Fortran is undefined behaviour: https://software.intel.com/en-us/forums/intel-fortran-compiler/topic/517808 , top reply.

I think we may need to autogenerate a cpp file in addition to the `occa_kernel_m.f90` file so we don't try to pass varargs across the Fortran-C++ boundary. Proof of concept (with functions lazily yet manually inserted in the place which was the least work) can be found at https://github.com/SFrijters/occa/tree/intel-vararg-poc . This version of the code fixes the problem for me (the max number of args used in the examples is 5).

@awehrfritz What do you think?

SFrijters:
This is still an issue though; I can provide a hardcoded fix for up to N kernel arguments, or the code gen scripts should be extended.

Issue #369, opened 2020-06-08
Internal Compiler Error with Intel compiler for Fortran example(s)


SFrijters:
Using the CMake build:

```
ifort (IFORT) 19.0.5.281 20190815
Copyright (C) 1985-2019 Intel Corporation.  All rights reserved.

[...]

cd /<redacted>/occa/build/intel-debug/examples/fortran/01_add_vectors && /<redacted>/EasyBuild/software/impi/2018.5.288-iccifort-2019.5.281/bin64/mpiifort -DUSE_CMAKE -I/<redacted>/occa/include -I/<redacted>/occa/build/intel-debug/include -I/<redacted>/occa/build/intel-debug/mod  -warn all -fPIE   -c /<redacted>/occa/examples/fortran/01_add_vectors/main.f90 -o CMakeFiles/examples_fortran_01_add_vectors.dir/main.f90.o

[...]

010101_13941

catastrophic error: **Internal compiler error: internal abort** Please report this error along with the circumstances in which it occurred in a Software Problem Report.  Note: File and line given may not be explicit cause of this error.
compilation aborted for /<redacted>/occa/examples/fortran/01_add_vectors/main.f90 (code 1)
```
And similarly for 03_static_compilation. I think I've narrowed it down to the call to `C_F_str`, but I'm not sure where to go from there. @awehrfritz , have you seen anything like this before?

dmed256:
If you need a temporary fix, setting
```
OCCA_FORTRAN_ENABLED=0
```
should disable Fortran compilation

SFrijters:
Yeah, but I was explicitly adding the Fortran examples to CMake.

awehrfritz:
@SFrijters Sorry for the late reply. I haven't seen this particular error before, but I have seen similar ones. In most cases this was a bug somewhere in the compiler itself.

When I get such errors with other Fortran codes that I frequently work with, I typically try a different compiler version which so far always make the problem go away - I know not the best solution. :) I have not tried using the Intel compiler with OCCA yet.

SFrijters:
@awehrfritz thanks for letting me know. I have found a workaround for the examples that failed to build - I'm not in a position to change my blessed compiler version. I will add this fix to my PR to add the Fortran interface to the CMake build.

mostafa-najafiyazdi:
@SFrijters , so what was your workaround? I am having the same issue.

SFrijters:
[...]

Issue #368, opened 2020-06-08
Add Fortran code to CMake


SFrijters:
I am currently working on adding the Fortran code and examples to CMake. I have a PR in progress, but I've run into some issue, so this is kind of a placeholder issue so others don't start implementing this as well.

SFrijters:
Closed by #374 .

PR #374, opened 2020-06-18
Add Fortran interfaces/tests/examples to CMake and fix various compiler issues

SFrijters:
## Description

This PR is a kind of a catch all, but the commits are a little more decoupled.

The main goal is to add the recently added Fortran interface, examples, and tests to the CMake code, locked behind the new ENABLE_FORTRAN option (default = OFF), closing #368.

In the process of getting this to work various warnings and errors came up for various compilers, and these have been (mostly) addressed in this PR as well. However, it does not address #372 or #373 and only has a workaround for #369.

This also adds header files when installing occa.


CMakeLists.txt
```diff
@@ -45,13 +50,21 @@ endif()
 
 include(SetCompilerFlags)
 
+string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPER)
```
noelchalmers, line 53: Since we're referencing `CMAKE_BUILD_TYPE` here, we should be sure it's set somewhere above. I'd suggest adding the following:
```
if (NOT CMAKE_BUILD_TYPE)
  set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Build type" FORCE)
endif()
```
in case the user or outside project hasn't declared their build type.

CMakeLists.txt
```diff
@@ -223,19 +237,36 @@ endif(ENABLE_MPI)
 
 #Generate CompiledDefines from libraries we found
 configure_file(scripts/compiledDefinesTemplate.hpp.in ${OCCA_BUILD_DIR}/include/occa/defines/compiledDefines.hpp)
+install(FILES ${OCCA_BUILD_DIR}/include/occa/defines/compiledDefines.hpp DESTINATION include/occa/defines)
```
noelchalmers, line 240: I may be showing my lack of cmake knowledge here, but what does this (and line 269 below) actually do? As far as I can tell they just set install directories to what they default to anyway.

CMakeLists.txt
```diff
@@ -213,6 +225,8 @@ if (ENABLE_MPI)
     message("-- MPI include dirs: ${MPI_CXX_INCLUDE_DIRS}")
     message("-- MPI libraries:    ${MPI_CXX_LIBRARIES}")
     message("-- MPI link flags:   ${MPI_CXX_LINK_FLAGS}")
+    message("-- MPI executable:   ${MPIEXEC_EXECUTABLE}")
+    message("-- MPI numproc flag: ${MPIEXEC_NUMPROC_FLAG}")
 
     target_include_directories(libocca PRIVATE ${MPI_CXX_INCLUDE_DIRS})
     target_link_libraries(libocca PRIVATE ${MPI_CXX_LIBRARIES})
```
noelchalmers, line 232: Since we're not explicitly passing the MPI includes and libs to the MPI examples in this PR, we should likely change these PRIVATE includes and libs to PUBLIC so they are propagated to targets. Otherwise theres a change the mpi.h and mpi libs aren't found in the examples.

CMakeLists.txt
SFrijters: Good point, will add this.

CMakeLists.txt
SFrijters: It tells CMake to actually install (copy) the referenced files/directories at install time. I think the library is something that is installed by default, but the (public) headers are not.
But looking at the recent CMake documentation it looks like there's a nicer way to do it, so I'll tweak this a bit.

CMakeLists.txt
SFrijters: This has not been a problem for my builds so far, but better safe than sorry.

CMakeLists.txt
SFrijters: Done.

CMakeLists.txt
SFrijters: Done.

CMakeLists.txt
SFrijters: Hm, the nicer way I thought about does not really work. Looking at some other examples online something similar is done as above. But I'm very much open to suggestions about a better / more modern way of doing it.

noelchalmers: Sorry, I'm still not really seeing what this does. In particular, after building I see only include/occa/defines/compiedDefines.hpp in my build folder. Moreover, commenting out this line(and line 273 now) seems to make no difference in the contents of the build folder, or with the build itself. Are these lines even needed?

noelchalmers: nevermind, I think I figured it out. Messaged in slack.

tests/src/lang/parser/statementLoading.cpp
```diff
@@ -1,5 +1,10 @@
 #include "utils.hpp"
 
+#ifdef __PGI
```
dmed256, line 3: What's the warning here?

SFrijters: I've split up my commits a bit more to more easily distinguish between the different fixes, so it should be easier to review the Fortran part.

The PGI warning I suppress here is (if I recall correctly, my development server is current undergoing maintenance so I can't check right now) that PGI is not 'fooled' by some of the `(void) statement` calls and complains about unused variable or variable is set but not used.

dmed256: Oh interesting. There's a `ignoreResult` method we use to void compiler errors. Is it coming from there?
https://github.com/libocca/occa/blob/f6333f2c808ddd29399b1f7ea0def461050d9d0c/include/occa/tools/misc.hpp#L11

SFrijters: If I don't put the pragma I get

```
"<redacted>/occa/tests/src/lang/parser/statementLoading.cpp
          ", line 468: warning #550-D: variable "statement" was set but never
          used
    statement_t *statement;
                 ^

"<redacted>/occa/tests/src/lang/parser/statementLoading.cpp
          ", line 628: warning #550-D: variable "statement" was set but never
          used
    statement_t *statement;
                 ^
```

Didn't know about the `ignoreResult` function - looks like I can use that instead.

SFrijters: Spliced it into the PGI commit instead of the pragma.

SFrijters:
@awehrfritz I'm not too worried about the fix you linked, those are just complaints about possibly uninitialized variables. But note that this PR doesn't address #372 , which in my opinion is a more important problem. If you have limited time I think it would be good to prioritize that one.

awehrfritz:
Ups, wrong commit in my comment, here the correct commit: https://github.com/libocca/occa/pull/374/commits/16f060281efcd5ad72ae98c8498f4840638f7dcc

SFrijters:
Ah, yes, that one's a bit worse. But it's an internal compiler error, so I'm not sure it's a bug in your Fortran code at all. At least nothing bad happens / can happen at runtime.

PR #386, opened 2020-07-13
Add size_t and ptrdiff_t to builtin types

SFrijters:
## Description

Currently the parser/preprocessor doesn't handle the `size_t` and `ptrdiff_t` types. I added them in the same places where `wchar_t` was handled to follow the example of another type that doesn't do much.

Should I add/extend a test for this, if this is a sensible addition at all?

### DIFF
--- a/include/occa/lang/builtins/types.hpp
+++ b/include/occa/lang/builtins/types.hpp
@@ -48,6 +48,10 @@ namespace occa {
     extern const primitive_t void_;
     extern const primitive_t auto_;
 
+    // TODO: Auto-generate type aliases for common C types
+    extern const primitive_t size_t_;
+    extern const primitive_t ptrdiff_t_;
+
     // OKL Primitives
     extern const primitive_t uchar2;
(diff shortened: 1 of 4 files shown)


SFrijters:
Well, CI gave me a hint that I should indeed at some test code path...

include/occa/lang/builtins/types.hpp
```diff
@@ -47,6 +47,8 @@ namespace occa {
     extern const primitive_t double_;
     extern const primitive_t void_;
     extern const primitive_t auto_;
+    extern const primitive_t size_t_;
```
dmed256, line 50: Here

src/lang/builtins/types.cpp
```diff
@@ -44,6 +44,8 @@ namespace occa {
     const primitive_t double_       ("double");
     const primitive_t void_         ("void");
     const primitive_t auto_         ("auto");
+    const primitive_t size_t_       ("size_t");
```
dmed256, line 47: Here

PR #395, opened 2020-08-11
Sys - add PPC for compiler flag detection

### LINKED ISSUE #394: PowerPC CPP Flags
I believe that commit 793a2a8fcc031f4e2b0ed6c21d08c0b5f8f95590 needs to be applied to [sys.cpp](https://github.com/libocca/occa/blob/b6d21d58dd445cd6ca47c85938449ba079ecf273/src/tools/sys.cpp) as well.

jeremylt:
## Description

@dmed256, I think this *might* fix the libCEED + OCCA PowerPC issue we see on our Travis, but I don't know enough about OCCA to know if there are other places where similar changes need to occur.

I didn't see your yml, but I know in the past you used Travis. If you're still using Travis,you can now test on various hardware: https://docs.travis-ci.com/user/multi-cpu-architectures/

Closes #394

### DIFF
--- a/include/occa/tools/sys.hpp
+++ b/include/occa/tools/sys.hpp
@@ -26,7 +26,8 @@ namespace occa {
       static const int b_HP           = 6;
       static const int b_VisualStudio = 7;
       static const int b_Cray         = 8;
-      static const int b_max          = 9;
+      static const int b_PPC          = 9;
+      static const int b_max          = 10;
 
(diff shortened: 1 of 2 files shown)


jeremylt:
Confirmed - this fixes the libCEED + OCCA PowerPC issue

jeremylt:
Though, now OCCA + PowerPC on Travis disagrees about if OpenMP is available.

PR #404, opened 2020-10-04
[Lang] Add statementArray methods and switch over CUDA

dmed256:
## Description

- Adds `statementArray` to make code transformation cleaner and safer
- Moves logic for fetching inner statements to `statement_t`
- Moves logic for fetching and replacing exprNodes to `statement_t`


dmed256:
Tested locally + on libP (Thank you @tcew!)

PR #405, opened 2020-10-22
[CMake] Fix intermittent errors when building Fortran examples/tests

SFrijters:
## Description

When building the Fortran examples/tests, the parallel build process would try to write to the same `.mod` files.
Don't force them to be put in the same directory, and CMake should keep them in separate directories per executable.
My builds have been stable since I introduced this change locally.

Also, opt/build would still show up as untracked files if they are symlinks, which is annoying.


SFrijters:
How does any of this affect code coverage?

dmed256:
> How does any of this affect code coverage?

It's sometimes flaky

I'm not 100% sure what the changes are since I'm not familiar with CMake. However, we can always add it back or look into it when we want to release the Fortran API as stable.

Thanks for the PR!

PR #413, opened 2020-10-27
[CMake] Fix changed subdirs in tests

SFrijters:
## Description

With recent commits a directory was removed and another one renamed, and this fixes CMake to use the right dirs again.


dmed256 (approved):
:fire:

PR #414, opened 2020-10-27
[Fix] shadowing warnings

SFrijters:
## Description

The Intel compiler complains about shadowing some variables. This PR fixes them in a way similar to 3e56817e9ab141068853130caab88a7263273b08 (adding underscores).


dmed256 (approved):
:fire:

PR #421, opened 2020-11-09
Make build system understand new hipconfig output

dmcdougall:
From ROCm 4.0, hipconfig --platform will output 'amd' instead of 'hcc'.
This change simply proactively mitigates a potential surprise where OCCA
thinks it couldn't find ROCm.


dmed256 (approved):
Thank you for the PR! :fire:

PR #424, opened 2020-11-26
[Lang] Add barriers only after last inner block in each kernel, skipping last inner block

### LINKED ISSUE #205: Add @barrier at the beginning of loops
When having `@inner` loops inside for-loops, we're missing a `@barrier` insertion at the beginning
```cpp
for (...) {
  for (...; @inner) {}
  for (...; @inner) {}
}
```
&darr;
```cpp
for (...) {
  @barrier() // <-- Missing!
  for (...; @inner) {}
  @barrier()
  for (...; @inner) {}
}
```

noelchalmers:
## Description

Fixes a bug where inner for loops were being replaced while checking if they were the outermost inner loop. Also adds one additional check for whether the inner block is the last in a kernel and skips adding a barrier if so.

Fixes #205

### DIFF
--- a/include/occa/lang/modes/withLauncher.hpp
+++ b/include/occa/lang/modes/withLauncher.hpp
@@ -44,6 +44,8 @@ namespace occa {
 
         bool isLastInnerLoop(forStatement &forSmnt);
 
+        bool isInsideLoop(forStatement &forSmnt);
+
         void setKernelLaunch(functionDeclStatement &kernelSmnt,
                              forStatement &forSmnt,
                              const int kernelIndex);
--- a/include/occa/lang/modes/withLauncher.hpp
+++ b/include/occa/lang/modes/withLauncher.hpp
@@ -42,6 +42,8 @@ namespace occa {
         bool isOuterMostOklLoop(forStatement &forSmnt,
                                 const std::string &attr);
 
+        bool isLastInnerLoop(forStatement &forSmnt);
+
         void setKernelLaunch(functionDeclStatement &kernelSmnt,
                              forStatement &forSmnt,
                              const int kernelIndex);
--- a/src/lang/modes/withLauncher.cpp
+++ b/src/lang/modes/withLauncher.cpp
@@ -446,20 +446,24 @@ namespace occa {
                 replaceOccaFor(outerSmnt);
               });
 
-        const bool applyBarriers = usesBarriers();
+        if (usesBarriers()) {
+          statementArray::from(kernelSmnt)
+              .flatFilterByAttribute("inner")
+              .filterByStatementType(statementType::for_)
+              .forEach([&](statement_t *smnt) {
+                  forStatement &innerSmnt = (forStatement&) *smnt;
+
+                  // TODO 1.1: Only apply barriers when needed in the last inner-loop
+                  if (isOuterMostInnerLoop(innerSmnt) )
+                    addBarriersAfterInnerLoop(innerSmnt);
+                });
+        }
 
         statementArray::from(kernelSmnt)
             .flatFilterByAttribute("inner")
             .filterByStatementType(statementType::for_)
             .forEach([&](statement_t *smnt) {
                 forStatement &innerSmnt = (forStatement&) *smnt;
-
-                // TODO 1.1: Only apply barriers when needed in the last inner-loop
-                if (applyBarriers &&
-                    isOuterMostInnerLoop(innerSmnt)) {
-                  addBarriersAfterInnerLoop(innerSmnt);
-                }
-
                 replaceOccaFor(innerSmnt);
               });
       }
--- a/src/lang/modes/withLauncher.cpp
+++ b/src/lang/modes/withLauncher.cpp
@@ -139,13 +139,11 @@ namespace occa {
 
       bool withLauncher::isOuterMostOklLoop(forStatement &forSmnt,
                                             const std::string &attr) {
-        statement_t *smnt = forSmnt.up;
-        while (smnt) {
-          if ((smnt->type() & statementType::for_)
-              && smnt->hasAttribute(attr)) {
+        for (auto &parentSmnt : forSmnt.getParentPath()) {
+          if (parentSmnt->type() & statementType::for_
+              && parentSmnt->hasAttribute(attr)) {
             return false;
           }
-          smnt = smnt->up;
         }
         return true;
       }
@@ -159,6 +157,15 @@ namespace occa {
           }
         }
         return true;
+      }
+
+      bool withLauncher::isInsideLoop(forStatement &forSmnt) {
+        for (auto &parentSmnt : forSmnt.getParentPath()) {
+          if (parentSmnt->type() & (statementType::for_ | statementType::while_)) {
+            return true;
+          }
+        }
+        return false;
       }
 
       void withLauncher::setKernelLaunch(functionDeclStatement &kernelSmnt,
@@ -465,7 +472,8 @@ namespace occa {
                   forStatement &innerSmnt = (forStatement&) *smnt;
 
                   //Only apply barriers when needed in the last inner-loop
-                  if (isOuterMostInnerLoop(innerSmnt) && !isLastInnerLoop(innerSmnt))
+                  if (isOuterMostInnerLoop(innerSmnt)
+                      && (!isLastInnerLoop(innerSmnt) || isInsideLoop(innerSmnt)))
                     addBarriersAfterInnerLoop(innerSmnt);
                 });
         }
--- a/src/lang/modes/withLauncher.cpp
+++ b/src/lang/modes/withLauncher.cpp
@@ -150,6 +150,17 @@ namespace occa {
         return true;
       }
 
+      bool withLauncher::isLastInnerLoop(forStatement &forSmnt) {
+        blockStatement &parent = *(forSmnt.up);
+        for(int smntIndex = forSmnt.childIndex()+1; smntIndex<parent.size(); smntIndex++) {
+          if ((parent[smntIndex]->type() & statementType::for_)
+              && parent[smntIndex]->hasAttribute("inner")) {
+            return false;
+          }
+        }
+        return true;
+      }
+
       void withLauncher::setKernelLaunch(functionDeclStatement &kernelSmnt,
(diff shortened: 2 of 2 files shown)


dmed256:
@noelchalmers I think we need to check if the statement is in a loop before removing the `@barrier`. Example:
```cpp
for (int n = 0; n < N; ++n) {
  for (int b = 0; b < B; ++b: @outer) {
    for (int i = 0; b < I; ++i: @inner) {
      // Inner 1
    }
    for (int i = 0; b < I; ++i: @inner) {
      // Inner 2
    }
  }
}
```
->
```cpp
for (int n = 0; n < N; ++n) {
  [A]
  // Inner 1
  @barrier
  // Inner 2
  [B]
}
```
in which we want a `@barrier` either in `[A]` or `[B]`. In the current code, it should be setting a second `@barrier` at `[B]`.

I like your change If we also add a check whether any parent statement is a non-okl loop. Thanks for making this PR!

noelchalmers:
Good catch! There should indeed be a barrier if there is an enclosing loop anywhere. I'll take a look at adding that check.

noelchalmers:
Thinking about this some more... The incorrect removal of the barrier is only a problem when removing the last barrier, right? All other times there should be a barrier following an inner block. 

So it seems like the last barrier can be removed, _unless_ it is enclosed in a non-okl loop before the outer loops. In your example, that barrier is actually safe to remove, as the outer loop is transformed into a kernel call. Whereas:
```c++
for (int b = 0; b < B; ++b: @outer) {
    for (int i = 0; b < I; ++i: @inner) {
      // Inner 1
    }
    ->barrier
    for (int n = 0; n < N; ++n) {
       for (int i = 0; b < I; ++i: @inner) {
         // Inner 2
       }
       ->barrier
    }
}
```
it is not safe to remove the last barrier after the second inner block. Unless I'm missing something (is it legal to have non-okl loops between outer(1) and outer(0) loops?), it should be sufficient to check if the last inner block is enclosed in a non-okl loop before it is enclosed in an outer loop. Do I have this right?

dmed256:
> In your example, that barrier is actually safe to remove, as the outer loop is transformed into a kernel call

Oops good catch, this is what generates the loop inside the kernel:
```
for (int b = 0; b < B; ++b: @outer) {
  for (int n = 0; n < N; ++n) {
    for (int i = 0; b < I; ++i: @inner) {
      // Inner 1
    }
    -> @barrier
    for (int i = 0; b < I; ++i: @inner) {
      // Inner 2
    }
    -> @barrier
  }
}
```

>  It should be sufficient to check if the last inner block is enclosed in a non-okl loop before it is enclosed in an outer loop. Do I have this right?

Super nested example:
```cpp
for (@outer) {
  for {
    -> @barrier         <-- first @Inner loop but needs an @barrier
                            because of the for-loop
    for(@inner) {}
    for {
      -> @barrier       <- Not first
      for(@inner) {}
      -> @barrier       <- Not first
      for(@inner) {}
    }
    for {
      -> @barrier       <- Not first
      for(@inner) {}
      for {
        -> @barrier     <- Not first
        for(@inner) {}
        -> @barrier     <- Not first
        for(@inner) {}
      }
      -> @barrier       <- Not first
      for(@inner) {}
    }
  }
}
```

A bit counter-intuitive but It might be easier if we insert the barriers before a given `@inner` loop instead of afterwards. If we group all the `@inner` loops by `@outer` loop, it looks like this should be the check we make:
```
if (isNotFirst || insideNonOklLoop) {
  up->insertBefore(innerLoop, barrier);
}
```
What do you think?

noelchalmers:
Adding barriers before the `@inner` loops seems equally easy as adding them after. After all, the analogous check would just be 
```
if (isNotLast || insideNonOklLoop) {
  up->insertAfter(innerLoop, barrier);
}
```
The super nested example then becomes:
```c++
for (@outer) {
  for {
    for(@inner) {}
    -> @barrier         <- Not last
    for {
      for(@inner) {}
      -> @barrier       <- Not last
      for(@inner) {}
      -> @barrier       <- Not last
    }
    for {
      for(@inner) {}
      -> @barrier       <- Not last
      for {
        for(@inner) {}
        -> @barrier     <- Not last
        for(@inner) {}
        -> @barrier     <- Not last
      }
      for(@inner) {}
      -> @barrier       <-- last @Inner loop but needs an @barrier
                                     because of the for-loop
    }
  }
}
```
I'm not sure if there's a major difference between having a slightly unnecessary `@barrier` at the beginning of a for-loop iteration at the start of a kernel, or one after the last for-loop iteration at the end. So it's not clear if we should favor adding `@barrier`s before or after inner blocks. 

The only thing that gives me pause is such an example:
```c++
for (@outer(1)) {
  for { //legal?
    for(@outer(0)) {
      for(@inner) {}
      -> @barrier       <- Not last
      for(@inner) {}
      -> @barrier       <- last @Inner loop, but @barrier because
                                    for-loop between @outer(0) and @outer(1)?
    }
  }
}
```
This is admittedly a weird kernel. I haven't tried something like this before.

dmed256:
> Adding barriers before the @inner loops seems equally easy as adding them after. After all, the analogous check would just be
> 
> ```
> if (isNotLast || insideNonOklLoop) {
>   up->insertAfter(innerLoop, barrier);
> }
> ```

Ah true, that seems good to me! :fire: 

> ```
> for (@outer(1)) {
>   for { //legal?
>     for(@outer(0)) {
> ```

It's legal as long as the okl-loops don't depend on the iterators. It's nice if we're iterating by `x`, `y`, and `z` but we want to dynamically change the attributes at run-time/okl-compile-time to set `@outer` on some of the 3 loops.

noelchalmers:
Ok, in that case since the kernel boundary will be at the outermost `@outer` loop, the check becomes a slightly-more-convoluted
```
if (isNotLast || insideNonOklLoopBeforeOutermostOkl) {
  up->insertAfter(innerLoop, barrier);
}
```
I'll look at how I can implement such a thing.

dmed256:
I don't think we need to check where the non-okl loop is as long as it's outside of the inner-most `@inner`-loop :thinking:

noelchalmers:
But is your first example not a case where we shouldn't add the `@barrier`? Namely, if the loop appears after the kernel boundary? E.g.
```c++
for {   
  for(@outer) {
    for(@inner) {}
    -> @barrier       <- Not last
    for(@inner) {}
                      <- last @Inner loop, but no @barrier because
                          for-loop is outside kernel boundary 
  }
}
```
Or are you saying that we do not have to check where the loop is because at the point we are adding `@barriers`, the kernels are already split, and we won't find that outermost for-loop?

noelchalmers:
That last commit should implement the check we discussed.

dmed256:
I tried it out with these 4 cases and it looks like it's adding an extra `@barrier` to the first case

<details>
  <summary>barrierTest.okl</summary>

  ```cpp
  @kernel void barrierTestKernel() {
    for (int i = 0; i < 10; ++i; @outer) {
      @shared int sharedPtr[2];
  
      for (int j = 0; j < 10; ++j; @inner) {
        sharedPtr[0] = 1;
      }
      // ^ @barrier?
      for (int j = 0; j < 10; ++j; @inner) {
        sharedPtr[0] = 1;
      }
    }
  
    for (int i = 0; i < 10; ++i; @outer) {
      @shared int sharedPtr[2];
  
      while (true) {
        for (int j = 0; j < 10; ++j; @inner) {
          sharedPtr[0] = 1;
        }
        // ^ @barrier?
        for (int j = 0; j < 10; ++j; @inner) {
          sharedPtr[0] = 1;
        }
        // ^ @barrier?
      }
    }
  
    for (int i = 0; i < 10; ++i; @outer) {
      @shared int sharedPtr[2];
  
      while (true) {
        for (int j = 0; j < 10; ++j; @inner) {
        }
        for (int j = 0; j < 10; ++j; @inner) {
          sharedPtr[0] = 1;
        }
        // ^ @barrier?
      }
    }
  
    for (int i = 0; i < 10; ++i; @outer) {
      @shared int sharedPtr[2];
  
      while (true) {
        for (int j = 0; j < 10; ++j; @inner) {
          sharedPtr[0] = 1;
        }
        // ^ @barrier?
        for (int j = 0; j < 10; ++j; @inner) {
        }
      }
    }
  }
  ```
  
</details>
  
<details>
  <summary>output.cu</summary>
  
  ```cpp
  extern "C" __global__ void _occa_barrierTestKernel_0() {
    {
      int i = 0 + blockIdx.x;
      __shared__ int sharedPtr[2];
      {
        int j = 0 + threadIdx.x;
        sharedPtr[0] = 1;
      }
       __syncthreads();
      // ^ @barrier?
      {
        int j = 0 + threadIdx.x;
        sharedPtr[0] = 1;
      }
       __syncthreads();
    }
  }
  
  extern "C" __global__ void _occa_barrierTestKernel_1() {
    {
      int i = 0 + blockIdx.x;
      __shared__ int sharedPtr[2];
      while (true) {
        {
          int j = 0 + threadIdx.x;
          sharedPtr[0] = 1;
        }
         __syncthreads();
        // ^ @barrier?
        {
          int j = 0 + threadIdx.x;
          sharedPtr[0] = 1;
        }
         __syncthreads();
        // ^ @barrier?
      }
    }
  }
  
  extern "C" __global__ void _occa_barrierTestKernel_2() {
    {
      int i = 0 + blockIdx.x;
      __shared__ int sharedPtr[2];
      while (true) {
        {
          int j = 0 + threadIdx.x;
        }
        {
          int j = 0 + threadIdx.x;
          sharedPtr[0] = 1;
        }
         __syncthreads();
        // ^ @barrier?
      }
    }
  }
  
  extern "C" __global__ void _occa_barrierTestKernel_3() {
    {
      int i = 0 + blockIdx.x;
      __shared__ int sharedPtr[2];
      while (true) {
        {
          int j = 0 + threadIdx.x;
          sharedPtr[0] = 1;
        }
         __syncthreads();
        // ^ @barrier?
        {
          int j = 0 + threadIdx.x;
        }
      }
    }
  }
  ```
  
</details>

dmed256:
Chatted offline and this PR looks like it has the intended changes. Something must have been messed up with my 2 OCCA clones :sweat_smile:

Thanks for making the changes!

PR #426, opened 2020-11-26
[HIP] Avoid checking the error code from hipGetDeviceCount

noelchalmers:
Since hipGetDeviceCount will throw a hipErrorNoDevice error when no devices are present, it's better to not check the error code so that `occa info` does not fail on GPU-less headnodes.


dmed256:
Nice, makes sense!

PR #427, opened 2020-12-01
[Modes] Adding a getDeviceCount API to query number of available devices in an enabled mode

noelchalmers:


dmed256:
so_beautiful

PR #429, opened 2020-12-04
Use arch name property to determine gpu arch

dmcdougall:
HIP added a gcnArchName member to the hipDeviceProps_t struct.  This is
a char array and is more suitable for determining the name of the
architecture to target.  The older gcnArch is an integer and is a little
more brittle.

### DIFF
--- a/src/occa/internal/modes/hip/polyfill.hpp
+++ b/src/occa/internal/modes/hip/polyfill.hpp
@@ -37,6 +37,7 @@ namespace occa {
     size_t totalGlobalMem;
     int maxThreadsPerBlock;
     int gcnArch;
+    char gcnArchName[256];
     int major;
     int minor;
 
--- a/src/occa/internal/modes/hip/utils.cpp
+++ b/src/occa/internal/modes/hip/utils.cpp
@@ -58,7 +58,7 @@ namespace occa {
                      hipGetDeviceProperties(&hipProps, deviceId));
 
       if (hipProps.gcnArch) {
-        return "gfx" + toString(hipProps.gcnArch);
+        return hipProps.gcnArchName;
       }
       std::string sm = "sm_";
       sm += toString(
--- a/src/occa/internal/modes/hip/utils.cpp
+++ b/src/occa/internal/modes/hip/utils.cpp
@@ -57,9 +57,14 @@ namespace occa {
       OCCA_HIP_ERROR("Getting HIP device properties",
                      hipGetDeviceProperties(&hipProps, deviceId));
 
-      if (hipProps.gcnArch) {
+      if (hipProps.gcnArch) { // AMD or NVIDIA
+#if HIP_VERSION >= 306
         return hipProps.gcnArchName;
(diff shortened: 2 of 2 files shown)


dmcdougall:
Note: this will not work for ROCm versions 3.5 or older.  Let me know if that's unacceptable.

dmcdougall:
Hmmm.  I don't know how to triage that macOS failure.  I presume you have ROCm headers installed on there.  I'm not sure what you expect from ROCm on a Mac.  The struct member I am targeting was made available in ROCm 3.6.

noelchalmers:
@dmcdougall The `hipDeviceProp_t` is actually faked inside the `occa` namespace when the HIP mode is note enabled. You'll have to add in the corresponding entry here: https://github.com/libocca/occa/blob/6c98e86c10b0dab514845abd87800b8623d62ebc/include/occa/modes/hip/polyfill.hpp#L30

dmcdougall:
I am happy to do that, but I'd argue there should be any HIP paths triggered at all when HIP mode is disabled.

src/modes/hip/utils.cpp
```diff
@@ -59,7 +59,7 @@ namespace occa {
                      hipGetDeviceProperties(&hipProps, deviceId));
 
       if (hipProps.gcnArch) {
-        return "gfx" + toString(hipProps.gcnArch);
+        return hipProps.gcnArchName;
```
dmed256, line 62: @dmcdougall Thanks for adding this, sorry for the delay in checking PRs! Can we initialize `gcnArchName` to some constant value like 
```cpp
static const std::string UNSET_VALUE = "OCCA_UNSET_VALUE";
...
  // Initialize gncArchName
  memcpy(gcnArchName, UNSET_VALUE.c_str(), UNSET_VALUE.size() + 1);
...
  // Leverage the arch name if possible (version > 3.5)
  if (strcmp(gcnArchName, UNSET_VALUE.c_str())) {
    return hipProps.gcnArchName;
  } else {
    return "gfx" + toString(hipProps.gcnArch);
  }
```

dmcdougall: I accomplished the same thing with a version check.  Is this acceptable?

dmed256:
```cpp
      if (hipProps.gcnArch) { // AMD or NVIDIA
#if HIP_VERSION >= 306
        return hipProps.gcnArchName;
#else
        return "gfx" + toString(hipProps.gcnArch);
#endif
      }
```
@dmcdougall Do we need to check that `gcnArchName` is set or will it always be set?

dmed256: Looks like it's safe from the tensor flow backend, good enough for me :)
image

dmcdougall: If it's not set, it's a ROCm bug, not an OCCA bug.  It will be set for new enough ROCm versions.

dmcdougall:
Thanks friendo!

PR #433, opened 2020-12-30
Refactor/split public and internal api

dmed256:
## Description

First big split on the public vs internal API


dmed256:
Manually merged to avoid squashing

PR #445, opened 2021-01-04
[Fix] Use correct new path for expermental mpi.hpp

SFrijters:
After the recent reshuffle these mpi headers were moved, but not corrected at their use sites.


dmed256:
Thanks for the fix!

PR #446, opened 2021-01-04
[Fix] Add exception for one Fortran test that doesn't fit the standard CMake functions

SFrijters:
## Description

One of the Fortran tests is weird in that it uses a cpp file as a library.
This file should *not* be used to build a standalone executable, but built as a library and then linked to the one test that uses it.

Perhaps there is a nicer way to fix this (the current commit is a bit ad-hoc), but as it stands it spits out linker errors if I build with `ENABLE_FORTRAN = ON` so it does need some attention.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -21,6 +21,10 @@ function(add_occa_test test_source)
   target_include_directories(${cmake_test_target} PRIVATE
     $<BUILD_INTERFACE:${OCCA_SOURCE_DIR}/src>)
 
+  if (${test_source} MATCHES "typedefs.f90")
+    target_link_libraries(${cmake_test_target} libtypedefs_helper libocca ${CMAKE_THREAD_LIBS_INIT} ${CMAKE_DL_LIBS})
+  endif()
+
   add_test(
     NAME ${cmake_test_target}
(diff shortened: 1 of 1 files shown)


dmed256:
Ah, maybe we should update the CMake test to build Fortran as well. Thank you for the second fix of the day!
image

SFrijters:
You're welcome, just trying to get our own Jenkins in the blue again. Thanks for the quick merges.

Regarding Fortran, #372 is still a thing that probably will not get exposed on your CI because it doesn't have the Intel compiler in there. I still have a proof of concept fix for that on our own fork.

dmed256:
The POC (https://github.com/SFrijters/occa/commit/fff430556cfa8eae846eca6b0c3c7ee2bf2fe6d4) looks good to me! Is this something we need to expose on the Fortran side as well or having it only on the `.h` and `.cpp` files is enough? I'm not familiar with Fortran

SFrijters:
Yeah, indeed you are right; I did a bad push that was missing a change on the Fortran side, the correct commit is 7a56b44e51a0dd80035e1ec4ed0082147f271110. The PoC only includes functions taking up to 5 args because that is what was needed for the examples, but it should really go into the codegen scripts?

PR #452, opened 2021-01-06
[Core] Make getDeviceCount syntax match new device::setup syntax

noelchalmers:
## Description

I noticed the device::setup syntax when passing a string has subtly changed from `device.setup("mode: 'Serial'")` to `device.setup("{mode: 'Serial'}")`. Forwarding that syntax change to the getDeviceCount API.


dmed256:
Nice catch, thanks for the fix!

PR #457, opened 2021-01-12
[OpenCL] Avoid checking the error code in getDeviceCountInPlatform

noelchalmers:
## Description
Similar to #426, AMD's OpenCL will throw a `CL_DEVICE_NOT_FOUND` on checking the deviceIDs on GPU-less head nodes. This PR avoids checking the error code returned when getting device counts, which avoids `occa info` failing on head nodes.


dmed256 (approved):
Thanks for the fix!

PR #466, opened 2021-01-20
[Fix] kernel header defines can be numbers

SFrijters:
## Description

Since #442 the kernels in our calling code refused to compile, because we set numerical defines.
It seems safe do allow this? It fixes our things at least.

### DIFF
--- a/src/core/kernel.cpp
+++ b/src/core/kernel.cpp
@@ -222,7 +222,8 @@ namespace occa {
 
     // Add defines
     for (const auto &entry : props["defines"].object()) {
-      if (entry.second.isString()) {
+      if (entry.second.isString() || entry.second.isNumber() ||
+        entry.second.isBool() || entry.second.isNull()) {
         kernelHeader += "#define ";
         kernelHeader += ' ';
         kernelHeader += entry.first;


dmed256:
Whoops, this was a mistake on a recent PR
</img>

Sorry about that, good catch!

src/core/kernel.cpp
```diff
@@ -222,7 +222,7 @@ namespace occa {
 
     // Add defines
     for (const auto &entry : props["defines"].object()) {
-      if (entry.second.isString()) {
```
dmed256, line 225: Can you revert to what the logic was before? 
```cpp
      kernelHeader += "#define ";
      kernelHeader += ' ';
      kernelHeader += entry.first;
      kernelHeader += ' ';
      kernelHeader += (std::string) entry.second;
      kernelHeader += '\n';
```

Or we can also check for `isBool()` and `isNull()`. I don't think we want to print out object or array types

SFrijters: How are booleans and nulls rendered as strings - will that be something that makes sense in a `#define`?
I can either remove the check completely, reverting to the previous logic, or add the full quartet, your call.

dmed256: Booleans will get stringified as `true` or `false`, and the null case as `NULL`. I like your approach since we explicitly limit the types we want to `#define`.

dmed256: Thanks for making the change, will merge after checks run 🎉

PR #468, opened 2021-01-20
[Fix] Shadowing warnings in functionDefinition.cpp

SFrijters:
## Description

Similarly to #414, the Intel compiler is again not happy about shadowing variables. I applied the same renaming as before.


dmed256:
So strange the compiler complains about shadowing in static methods 🤔 

Thanks again for the fix!

PR #483, opened 2021-02-10
Attempt to make kernelProperties / compilers more consistent

SFrijters:
I recently ran into an issue where some of our kernels were C++ and others were C.
This led to issues because assumptions were made based on the first kernel hit.

This PR is an attempt to improve that situation - it works for us, but I expect some discussion is needed to get this (or something like this) into the main code.

At the least, I think `const bool compilingCpp = (((int) kernelProps["compiler_language"]) == sys::language::CPP);` in the kernel part of the code is broken and should be `const bool compilingCpp = lowercase(compilerLanguage) != "c";` as in the device part.

But this PR is a bit more invasive and delegates the determination of the kernel settings fully to the `buildKernel` call.
Although it is mostly a copy-paste of the code that was inside the device part, it has some changes.
The code now first determines the language for that kernel and takes values from the environment based on that (following the original pattern of OCCA_env > json["env"] > env. This is also extended for plain C code.

Conceptually, it is not quite clear to me what compiler/compiler flags as a property on the device should represent. Is this just a convenient place to store 'defaults' per device, or is there a deeper meaning?


dmed256 (approved):

dmed256:
LGTM, feel free to merge if you're happy with the changes as they are. Sorry it took a while to review 😞

SFrijters:
No problem about the delay, I'm never blocked by my own PRs since we use them on our own fork first for testing before upstreaming.
Glad you like the changes - but I cannot merge them myself `"You’re not authorized to merge this pull request. "`.

PR #486, opened 2021-03-08
[HIP] Removing use of __HIP_ROCclr__ since it is marked as deprecated. Using more stable HIP_VERSION check

noelchalmers:
## Description

The `__HIP_ROCclr__` define we use to check for the newer hipcc syntax is marked for deprecation in a future ROCm. This PR switches to use a `HIP_VERSION` check instead, which is a more stable way to determine the underlying ROCm version.


dmcdougall:
Approved.

Issue #487, opened 2021-03-15
Wrong thread index


stgeke:
Current repo version seems to mess up threadIdx

Dummy kernel:
```
@kernel void foo(int N, @restrict float* u)
{
 for (int e = 0; e < N; e++; @outer) {
  for(int i = 0; i < 8; ++i; @inner) {
   for(int j = 0; j < 8; ++j; @inner) {
   }
  }
 }
}
```
occa translate -m CUDA produces:

```
extern "C" __global__ void _occa_foo_0(int N,
                                       float * __restrict__ u) {
  {
    int e = 0 + blockIdx.x;
    {
      int i = 0 + threadIdx.x; // correct is threadIdx.y
      {
        int j = 0 + threadIdx.x; 
      }
    }
  }
}
```

PR #489, opened 2021-03-15
Fixes issue #487 related to incorrect translation of `@inner` index.

krowe-alcf:
Issues #487 affects not only the CUDA backend, but also OpenCL, HIP, etc.

### DIFF
--- a/src/occa/internal/lang/modes/oklForStatement.cpp
+++ b/src/occa/internal/lang/modes/oklForStatement.cpp
@@ -388,8 +388,11 @@ namespace occa {
             }
           });
 
-        // Index is 1 less than count
-        return maxInnerCount ? maxInnerCount - 1 : 0;
+        // @outer index is 1 less than count
+        if(oklAttr_ == "outer")
+          maxInnerCount--;
+
+        return maxInnerCount;
       }
 
       statementArray oklForStatement::getOklLoopPath() {


src/occa/internal/lang/modes/oklForStatement.cpp
```diff
@@ -388,8 +388,11 @@ namespace occa {
             }
           });
 
-        // Index is 1 less than count
-        return maxInnerCount ? maxInnerCount - 1 : 0;
+        // @outer index is 1 less than count
+        if(oklAttr_ == "outer")
```
bl-ue, line 392: I'm not familiar with this project, but I'd like to ask: are you sure you check `oklAttr_` directly?  Maybe you could use `oklForStatement::isOuterLoop()` (declared a few lines above, [here](https://github.com/libocca/occa/blob/main/src/occa/internal/lang/modes/oklForStatement.cpp#L353-L355)):
```suggestion
        if (isOuterLoop())
```
But as I said I don't know this project well, so I might be wrong.  I do notice `oklAttr_` is passed to this function.

dmed256: Good callout, that is cleaner! Kris and I chatted offline and it looks like he found where the issue is coming from but it might not be the right solution. We might be skipping or including extra loop counts in the `forOklForLoopStatements` method. I'll put some time soon to look more into this method, but thank you both for the inputs!

dmed256:
Found the bug and patched it in PR #492. Thank you @krowe-alcf for finding the source of the issue!

Issue #449, opened 2021-01-06
Presumed race condition in caching


jedbrown:
In trying to address remaining concerns in https://github.com/CEED/libCEED/pull/688, I've been switching between occa main and 1.1.0 (the version that was working previously). I clean, remove `~/.occa`, and rebuild. Sometimes I get errors like below. Seems to be a race conditioning related to caching. We typically `make -j64 test`.

If this issue is hard to fix (or others may persist), how would you recommend isolating our build environments so that there is no chance of colliding with stale caches?

```
not ok 33 t521-operator /gpu/hip/occa stderr
# +terminate called after throwing an instance of 'occa::exception'
# +what():
# +---[ Error ]--------------------------------------------------------------------
# +File     : /home/gitlab-runner/builds/N8GCsqus/0/libceed/occa-1.1.0/src/tools/sys.cpp
# +Line     : 855
# +Function : dlopen
# +Message  : Error loading binary [5f0237e1aff76bcc/launcher_binary] with dlopen: /home/gitlab-runner/.occa/cache/5f0237e1aff76bcc/launcher_binary: file too short
# +Stack
# +19 occa::error(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char
> > const&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
# +18 occa::sys::dlopen(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, occa::io::lock_t const&)
# +17 occa::serial::device::buildKernelFromBinary(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_t
raits<char>, std::allocator<char> > const&, occa::properties const&, occa::lang::kernelMetadata_t&)
# +16 occa::serial::device::buildKernelFromBinary(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_t
raits<char>, std::allocator<char> > const&, occa::properties const&)
# +15 occa::serial::device::buildKernel(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char
>, std::allocator<char> > const&, occa::hash_t, occa::properties const&, bool)
[...]

PR #493, opened 2021-04-13
WIP Replace file locking with temp file + rename semantics

SFrijters:
## Description

A step towards fixing #449: remove all locking code and use a temporary file + rename instead.

Pros:
- We've been using these changes on our fork for a while now and it seems to work - I haven't seen any `file too short` errors in automated testing anymore and I haven't heard from users either.
- Removes a bunch of code.

Cons:
- The code as I wrote it is quite ugly. In particular, the creation of a temporary directory requires 'manual' cleanup via coding conventions (see my comments in `src/occa/internal/io/utils.cpp`). I found it a bit tricky to make a nicer interface because the filename is sometimes used by occa directly, and sometimes used as a compiler argument etc.
- I don't have a Windows implementation yet, and I don't have a system to test a solution on.
- This doesn't prevent simultaneous compilation of the same code by different processes, only possible errors caused thereby. It would be nice to not have concurrent compiler invocations that could be avoided.

But before I go deeper I would like to solicit feedback on the direction and any bad assumptions I may have made, regarding correctness and otherwise.


SFrijters:
@jedbrown Do you have an opinion on this solution? It's pretty battle-tested by now on our local fork.

I'll fix the compiler complaint soon.

src/occa/internal/io/utils.cpp
```diff
@@ -442,5 +442,65 @@ namespace occa {
...
+                 status == 0 || io::isFile(filenameOut));
+
+      /*
+        Assume the temporary file has been created using the tmpFilenameBelow
+        function, so we need to clean up the temporary directory as well.
+      */
+      const std::string tmpDirname = dirname(io::expandFilename(filenameIn));
+      status = rmdir(tmpDirname.c_str());
+      OCCA_ERROR("Failed to remove temporary directory [" << tmpDirname << "]: " << strerror(errno),
+                 status == 0);
```
jedbrown, line 502: If there's no way in which the directory gets populated with other files, maybe the directory doesn't need to be created (it's just more precious metadata on a parallel filesystem).

amikstcyr:
@noelchalmers  can you please help with the windows part of this?

src/occa/internal/io/utils.cpp
```diff
@@ -442,5 +442,65 @@ namespace occa {
...
+      char buffer[bufferSize + 1];
+      ::memcpy(buffer, template_.c_str(), bufferSize);
+      // Set null terminator
+      buffer[bufferSize] = '\0';
+      char *dirname = mkdtemp(buffer);
+      OCCA_ERROR("Failed to create temporary directory for template [" << template_ << "]: " << strerror(errno),
+                 dirname != NULL);
+      return std::string(dirname) + "/" + std::string(basename(expFilename));
+#else
+#error "TODO: What to use on Windows?"
```
jedbrown, line 475: > If new_filename exists, the behavior is implementation-defined. 
https://en.cppreference.com/w/cpp/io/c/rename

Perhaps choose this when available:
> existing non-directory file: new_p is first deleted, then, without allowing other processes to observe new_p as deleted, the pathname new_p is linked to the file and old_p is unlinked from the file. Write permissions are required to both the directory that contains old_p and the directory that contains new_p. 
https://en.cppreference.com/w/cpp/filesystem/rename

jedbrown (commented):
Sounds good. I can test with libCEED whenever you expect it to work.

src/occa/internal/io/utils.cpp
SFrijters: I would also prefer not to have a directory but a safe filename instead (to go into the base cache dir), but I didn't see a way to get one reliably. See my comment at the top of `tmpFilenameBelow`. If you have a better way I'll gladly rewrite this bit.

SFrijters:
I don't know what's up with the GitHub actions here... but `/usr/bin/bash: clang++-9: command not found` surely isn't the fault of my commits.

@jedbrown We've been using this code for a while now (basically since I opened the pull request) and it should work: I haven't seen any caching errors on our CI or heard complaints from users after introducing this change. So if you can test it as well that would be great.

src/occa/internal/io/utils.cpp
jedbrown: So if there was an active adversary, they could observe the directory being created and then plop stuff in there. I don't think the compiler output file can be made robust to that (because you can't make the compiler use `mkstemp` internally and report back the name), so I'd instead focus on creating file names that don't collide. A hash of the content (or a cheap non-repeating number) plus a per-process salt (which could be strong, as in https://en.cppreference.com/w/cpp/numeric/random/random_device) would be more secure and debuggable.

SFrijters:
Testing blocked by #499 .

src/occa/internal/io/utils.cpp
SFrijters: From the security perspective, this temp dir will always be below the occa cache dir, so any adversary could then probably also just populate the cache with malicious binaries directly?

Since this temporary file name is often used to capture the output I don't see how we can consistently use a hash based on that.

Naively I would try to maybe use something like generating a hash from (hashdir + pid + time + random number) to use as a filename?

src/occa/internal/io/utils.cpp
SFrijters: Rename is already in use for the actual renaming, the TODO was for generating the file/directory name.
But that will be platform agnostic if we don't rely on `mkdtemp` anymore.

src/occa/internal/io/utils.cpp
jedbrown: I mean that you could hash the input, such that the filename would reflect changes (or not), which might be useful when debugging. But anything with enough uniqueness would be fine.

src/occa/internal/io/utils.cpp
SFrijters: `std::filesystem` is C++17, so unless we want to change from C++11 I'll have to stick with `std::rename` for now.

src/occa/internal/io/utils.cpp
SFrijters: We are already inside a hashDir, which is based on the contents of the kernel file. So maybe we could also look at the contents of the intermediate files that we are using, but it sounds expensive?

The PID seems to be hard to come by in a platform-independent way, so I'm just experimenting with a hash based on time + random at the moment. Code compiles and runs, but I'm not actually stress-testing it at the moment.

jedbrown:
You'll need this to compile
```diff
diff --git i/src/occa/internal/modes/serial/device.cpp w/src/occa/internal/modes/serial/device.cpp
index b9303742..46337208 100644
--- i/src/occa/internal/modes/serial/device.cpp
+++ w/src/occa/internal/modes/serial/device.cpp
@@ -81,7 +81,7 @@ namespace occa {
       }
 
       if (!io::isFile(outputFile)) {
-        const std::string outputFileTmp = io::tmpFilenameBelow(outputFile);
+        const std::string outputFileTmp = io::tmpFilenameFor(outputFile);
         parser.writeToFile(outputFileTmp);
         io::renameTmpFile(outputFileTmp.c_str(), outputFile.c_str());
       }
```

I can't test without implementing some OCCA API changes in https://github.com/CEED/libCEED/pull/688. To save us time in the event @dmed256 can't finish his branch, is there a migration guide for what happened to `kernelBuilder::fromString` and such?

stgeke:
rename(2) guarantees atomicity (either the old or the new file exists) but not durability (data was written to disk).
To be on the safe side I suggest:

1. write to temporary name
2. fsync new file
3. rename temporary name to target
4. fsync directory containing target

jedbrown:
Hi @SFrijters. I see the blocker #499 has finally merged. Is this able to move forward now?

dmed256:
@SFrijters Used your solution in ##505, thank you for the PR!

PR #494, opened 2021-05-10
[Modes] DPC++ backend

kris-rowe:
## Description

This PR implements a new OCCA backend using [Data Parallel C++](https://docs.oneapi.com/versions/latest/dpcpp/index.html) (DPC++). 

### Highlights

- Using freely available DPC++ implementations, this backend can be run on CPUs, GPUs, and FPGAs.
- Memory allocation/deallocation is handled using the DPC++ [USM functions](https://docs.oneapi.com/versions/latest/dpcpp/usm.html), which are now part of the SYCL 2020 standard.
- OKL kernels are translated to lambda functions for transparency and performance purposes.
- `@outer` and `@inner` loop indices are mapped to work-group IDs and work-items, respectively &mdash; similar to OpenCL.

### Building

Currently only building with CMake is supported. 

### Coverage

During development the [Intel oneAPI DPC++ compiler](https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/dpc-compiler.html#gs.125cio) was used. Intel GPU hardware was used for testing, including the [JLSE Aurora testbeds](https://www.jlse.anl.gov/hardware-under-development/) at ALCF. In the near future testing will be expanded to included other DPC++ implementations and other vendor hardware.

### Acknowledgements

This work was completed jointly by ALCF and Intel, with contributions from

- Anoop Madhusoodhanan Prabha (Intel)
- Cedric Andreolli (Intel)
- Kris Rowe (ALCF)
- Phillipe Thierry (Intel)
- Saumil Patel (ALCF)


examples/cpp/01_add_vectors/addVectors.okl
```diff
@@ -2,7 +2,7 @@
                         const float *a,
                         const float *b,
                         float *ab) {
-  for (int i = 0; i < entries; ++i; @tile(16, @outer, @inner)) {
+  for (int i = 0; i < entries; ++i; @tile(4, @outer, @inner)) {
```
dmed256, line 5: Is there a reason we want to block by 4 instead of 16? This feels like a very tiny warp size for GPU modes

examples/cpp/06_shared_memory/reduction.okl
```diff
@@ -2,17 +2,15 @@
                                        const float *vec,
                                        float *blockSum) {
   // Partial reduction of vector using loop tiles of size block (power of 2)
-  for (int group = 0; group < ((entries + block - 1) / block); ++group; @outer) {
+  for (int group = 0; group < ((entries + block - 1) / block); ++group; @tile(16,@outer)) {
```
dmed256, line 5: Any reason this file changed?

examples/cpp/20_native_dpcpp_kernel/addVectors.cpp
```diff
@@ -0,0 +1,12 @@
+#include <CL/sycl.hpp>
+
+extern "C" void addVectors_it(::sycl::queue *q, ::sycl::nd_range<3> *ndrange, int &entries, int *oa, int *ob, int *oc)
```
dmed256, line 3: Can we keep the variables consistent with the other examples? Splitting the arguments in new lines can also make the expected arguments + ordering more readable

```suggestion
extern "C" void addVectors_it(::sycl::queue *q,
                              ::sycl::nd_range<3> *ndrange,
                              int &entries,
                              int *a,
                              int *b,
                              int *ab) {
```

examples/cpp/20_native_dpcpp_kernel/main.cpp
```diff
@@ -0,0 +1,109 @@
...
+  occa::memory o_a = device.malloc<int>(entries);
+  occa::memory o_b = device.malloc<int>(entries);
+  occa::memory o_ab = device.malloc<int>(entries);
+
+  // Compile a regular DPCPP kernel at run-time
+  occa::properties kernelProps;
+  kernelProps["okl/enabled"] = false;
+ 
+  occa::kernel addVectors = device.buildKernel("addVectors.cpp",
+                                               "addVectors_it",
```
dmed256, line 47: Any reason this method is named differently from other examples?
```suggestion
                                               "addVectors",
```

src/occa/internal/bin/occa.cpp
```diff
@@ -156,6 +157,8 @@ namespace occa {
         parser = new lang::okl::openclParser(kernelProps);
       } else if (mode == "metal") {
         parser = new lang::okl::metalParser(kernelProps);
+      }else if (mode == "dpcpp") {
```
dmed256, line 160: Nit:
```suggestion
      } else if (mode == "dpcpp") {
```

src/occa/internal/lang/builtins/types.cpp
```diff
@@ -89,5 +89,12 @@ namespace occa {
...
     const primitive_t double3       ("double3");
     const primitive_t double4       ("double4");
-  }
+    
+    // DPCPP primitives
+    const primitive_t syclQueue     ("sycl::queue");
+    const primitive_t syclNdRange   ("sycl::nd_range<3>");
+    const primitive_t syclNdItem    ("sycl::nd_item<3>");
+    const primitive_t syclHandler   ("sycl::handler");
+    const primitive_t syclAccessor  ("auto");
```
dmed256, line 98: I don't think these should be OKL primitive types. Any reason these are needed?

src/occa/internal/lang/expr/dpcppAccessorNode.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace occa
+{
+  namespace lang
+  {
+
+    dpcppAccessorNode::dpcppAccessorNode(token_t *token_,
+                                         const vartype_t &shared_type_,
+                                         const std::string &handler_name_)
+        : exprNode{token_}, shared_type{shared_type_}, handler_name{handler_name_}
```
dmed256, line 12: Nit: We aren't using the `{}` intializers, but it doesn't seem to impact anything either so might be ok to keep
```suggestion
        : exprNode(token_), shared_type(shared_type_), handler_name(handler_name_)
```

src/occa/internal/lang/expr/exprNode.cpp
```diff
@@ -60,6 +60,9 @@ namespace occa {
       const udim_t parentheses       = (((uint64_t) 1) << 35);
       const udim_t tuple             = (((uint64_t) 1) << 36);
       const udim_t cudaCall          = (((uint64_t) 1) << 37);
+      const udim_t lambda	           = (((uint64_t) 1) << 38);
+      const udim_t dpcppAccessor     = (((uint64_t) 1) << 39);
+      const udim_t dpcppAtomic       = (((uint64_t) 1) << 40);
```
dmed256, line 65: Nit:
```suggestion
      const udim_t lambda            = (((uint64_t) 1) << 38);
      const udim_t dpcppAccessor     = (((uint64_t) 1) << 39);
      const udim_t dpcppAtomic       = (((uint64_t) 1) << 40);
```

src/occa/internal/lang/modes/withLauncher.cpp
```diff
@@ -41,19 +43,24 @@ namespace occa {
         if (!success) return;
         setOklLoopIndices();
 
-        if (!success) return;
+        if (!success)
```
dmed256, line 46: Any reason for these changes?

src/occa/internal/lang/type/lambda.cpp
```diff
@@ -0,0 +1,126 @@
...
+      {
+      case capture_t::byValue:
+        pout << "=";
+        break;
+      case capture_t::byReference:
+        pout << "&";
+        break;
+      default:
+        pout << "???";
+        break;
```
dmed256, line 96: Is this required?

src/occa/internal/lang/type/pointer.cpp
```diff
@@ -2,18 +2,18 @@
...
+    //   qualifiers(other.qualifiers) {}
 
-    pointer_t& pointer_t::operator = (const pointer_t &other) {
-      qualifiers = other.qualifiers;
-      return *this;
-    }
+    // pointer_t& pointer_t::operator = (const pointer_t &other) {
+    //   qualifiers = other.qualifiers;
+    //   return *this;
+    // }
```
dmed256, line 16: Why was this commented out? If we don't need it we can remove it

src/occa/internal/modes/dpcpp/utils.cpp
```diff
@@ -0,0 +1,160 @@
...
+  namespace dpcpp
+  {
+    /* Returns true if any DPC++ device is enabled on the machine */
+    bool isEnabled()
+    {
+      auto device_list = ::sycl::device::get_devices();
+      // The SYCL standard states that, when no arguments are passed to
+      // device::get_devices (default argument is info::device_type::all),
+      // at least one device is returned: i.e. the SYCL host device. 
+      return (device_list.size() > 1);
```
dmed256, line 25: Should we check `> 0` instead?
```suggestion
      return (device_list.size() > 0);
```

dmed256 (commented):
Overall it seems good to me!

Global comments:
- Was The file `include/occa/core/.device.hpp.swp` meant to be committed?
- Not sure if an IDE auto-formatted a lot of these files but there are a few differences that were added such as:
    - No inlined `return` but it also didn't add `{}` when it split the lines
    - `{` in the next line even though we put them in the same line
    - Casting doesn't include spaces (e.g. `(int) a` &rarr; `(int)a`)
    - I guess we should add a C++ formatter if we want to keep code styling consistent 😞

examples/cpp/01_add_vectors/addVectors.okl
kris-rowe: This likely got changed during testing to ensure more than one work-group is launched since the total entries for the example is only 12. The tile size of 16 (or large) makes more sense, but then we should change the number of entries to 128 or larger also.

examples/cpp/06_shared_memory/reduction.okl
kris-rowe: I am not 100% sure why this was changed. It can be removed.

examples/cpp/20_native_dpcpp_kernel/main.cpp
kris-rowe: At some point there may have been a reason, but running the example with the name "addVectors" passes.

src/occa/internal/lang/type/pointer.cpp
kris-rowe: It can be removed.

src/occa/internal/modes/dpcpp/utils.cpp
kris-rowe: At least one device will always be returned (which is the host device); however, we have excluded this previously in the device registration. The host device can be useful for debugging, so lets change this to zero and I will update the device registration function to include the host device.

src/occa/internal/lang/modes/withLauncher.cpp
kris-rowe: Apologies, this is probably some auto formatting again.

src/occa/internal/lang/type/lambda.cpp
kris-rowe: We could default to `byValue` or `byReference` as an alternative.

dmed256 (approved):

tcew:
The MacOS gcc-9 test fails curiously with 

```./tests/run_examples: line 108: cd: /Users/runner/work/occa/occa/examples/c/03_inline_okl: No such file or directory```

I am not sure what is out of sync here.

Issue #379, opened 2020-06-25
Add support for non-blocking streams


stgeke:
Currently all streams are synced w.r.t. the NULL stream:
https://github.com/libocca/occa/blob/master/src/modes/cuda/device.cpp#L132

noelchalmers:
After #378 is patched, the only thing that could be in the NULL stream is blocking mem copies. Is a option non-implicitly-synchronizing NULL stream really required when the application developer can simply use async copies in other streams? What's the benefit here?

stgeke:
It would allow a user to overlap kernels in a non NULL stream with kernels in the NULL stream.

noelchalmers:
I'm aware, but my question is why should this be a separate option in OCCA's stream creation. The same behavior can be obtained by the application developer putting all work in non-NULL streams as usual. OCCA itself puts nothing in the NULL stream, aside from blocking memcpys, which can be avoided by the developer by using async copies.

stgeke:
It's just for convenience or if you're a lib developer which cannot control what happens in the upstream application. Just mimicking what e.g. the CUDA API exposes with cudaStreamNonBlocking.

PR #498, opened 2021-06-18
[#379][Stream]: Add non-blocking stream creation

deukhyun-cha:
## Description
- Add creating a non-blocking stream based on a stream property value "nonblocking"
- Current implementation applies to cuda and hip modes only
- An example is added to demonstrate the usage


noelchalmers:
I'm not sure I see the value-add of this. 

- This option only affects how streams would or would not overlap with work on the NULL stream. But the NULL stream isn't used for any kernels/async APIs in OCCA. Hence this behavior would only be usable in a user's code that is interop-ing OCCA with CUDA or HIP (modulo some blocking memcpys which are always avoidable). 
- Consequently, a more valuable and flexible contribution would probably be a `device::wrapStream` method, rather than additional options at stream creation. 
- That new method would be able to support OpenCL's commandQueue options as well. 

Just my $0.02 CND.

tcew:
Is that like $0.01 US?

amikstcyr:
Thank you for the feedback!

In our opinion, this PR seems:

1. to have roughly zero overhead cost,
2. to have no effect if the underlying API/lib is not syncing with some sort of base stream,
3. to be useful when interfacing with certain libraries,
4. to be ready & responds to a need/use case hence immediately useful.

We could work on your suggestion but not sure why doing so would block this particular PR.

PR #499, opened 2021-06-21
[Tests] Update GitHub build

SFrijters:
## Description

ubuntu-latest has moved beyond 20210517.1, from which point clang-9 is no longer included - 10/11/12 are available.

Drop clang 9, use clang 11 instead.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,17 +30,17 @@ jobs:
             OCCA_COVERAGE: 1
             useCMake: true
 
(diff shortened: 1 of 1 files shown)


SFrijters:
For reference https://github.com/actions/virtual-environments/commit/75f9131c187acdb0dcea8c39b16d64f6c482b866

SFrijters:
@dmed256 It would be nice if you (or another person with commit rights) could merge this (or suggest another fix - I'm happy to change to different compilers or image instead), since this blocks all CI at the moment.

PR #510, opened 2021-09-07
[OpenMP] Fix conflict with re-used stringstream in closure

SFrijters:
## Description

Previously this used the 'ss' stringstream both inside the closure and outside,
resulting in the first element of the compiler invocation (i.e. the compiler executable)
sometimes ending up in flag, instead of the actual OpenMP flag.


dmed256 (approved):
Nice find!

PR #518, opened 2021-09-18
[Bugfix][OpenCL] Update opencl timing to give meaningful results.

kris-rowe:
Currently using `device::tagStream` and `device::timeBetween` to time kernels with the OpenCL backend results in negative values.


dmed256 (approved):
🔥

PR #520, opened 2021-09-20
fsync kernel build output

### LINKED ISSUE #515: Build artifacts (e.g. binaries for kernel + launcher) are not dureable
It would be better to fsync() them including the directory containing them. Otherwise you may run into unexpected issues if different compute nodes share the same filesystem.

stgeke:
Fixes #515

### DIFF
--- a/src/occa/internal/io/utils.cpp
+++ b/src/occa/internal/io/utils.cpp
@@ -4,6 +4,7 @@
 #include <stddef.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <fcntl.h>
 
 #include <occa/defines.hpp>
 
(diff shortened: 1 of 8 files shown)


stgeke:
Please let me know what it takes to merge this PR.

kris-rowe:
Apologies for delay. The only thing I can see is that this may not be applicable in the Windows case, however adding some precompiler guards for now should likely be okay.

stgeke:
May we can use this on Windows:
https://docs.microsoft.com/en-gb/windows/win32/api/fileapi/nf-fileapi-flushfilebuffers?redirectedfrom=MSDN

nnnunnn:
My [Windows fork](https://github.com/nnnunnn/occa) uses: **int _commit(int fd);**

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/commit?view=msvc-160

From [enable_msvc - occa/internal/io/utils.cpp line 471](https://github.com/nnnunnn/occa/blob/2c77f219935846be3c6dee7fdd3cea5e349644c9/src/occa/internal/io/utils.cpp#L471):

```
#if (OCCA_OS & (OCCA_LINUX_OS | OCCA_MACOS_OS))
      fsync(fileno(fp));
#else
      _commit(fileno(fp));        // NBN: VC++ compiler
#endif
```

kris-rowe:
This PR has been updated in #638

PR #521, opened 2021-09-20
Add kernel prop kernelNameSuffix

stgeke:


stgeke:
Please let me know what it takes to merge this PR.

stgeke:
What's the rational not to include it? A user kernel suffix is very handy when you want to profile a kernel running in multiple configurations (e.g. FP32 and FP64).

noelchalmers:
I think Kris's point, which I agree with, is that this can be easily accomplished with a macro in the okl source.

PR #522, opened 2021-10-01
Add logging for high memory watermark.

MalachiTimothyPhillips:


MalachiTimothyPhillips:
Thanks for the unit test reminder!

PR #531, opened 2021-10-22
[OKL][Feature] `@max_inner_dims` attribute

### LINKED ISSUE #307: New kernel qualifier macro
It is often helpful to give the kernel compiler some extra information about the number of threads (work-items) in a thread-block (work-group). For instance we can give the HIP compiler an upper bound on the number of threads in a thread-block (say 1024) as follows:

```__launch_bounds__(1024) __global__ void fooKernel(...) { ... }```

In fact for the current HIP release it is unfortunately the case that this must be specified when the thread-block size exceeds 256 (see https://github.com/ROCm-Developer-Tools/HIP/issues/1310 )

CUDA also has the same attribute. There is also an extra argument to the launch bounds qualifier for minimum number of thread-blocks (https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#launch-bounds).

The kernel qualifiers in OpenCL are slightly different (see 6.7.2 of https://www.khronos.org/registry/OpenCL/specs/2.2/pdf/OpenCL_C.pdf )

**Proposal v1** - in an ideal world when we know the thread-block size at compile time OCCA will add kernel qualifiers for appropriate launch bounds (CUDA, HIP) or work-group size hint (OpenCL). 

**Proposal v2** - If it is too complicated to do Proposal v1, then it would neat to add an okl attribute for launch bounds @qualifier("inner sizes", B) where B could be a compiler define. This would be expanded to  __launch_bounds__(value of B) for CUDA/HIP or __attribute__((work_group_size_hint(value of B))) for OpenCL. Multi-dim variant would also be helpful.

### PR DESCRIPTION
**Pull request by kris-rowe**
## Description
Closes #307

Introduces a new loop attribute `@max_inner_dims(X, Y, Z)`, where `X`, `Y`, and `Z` are compile-time constants. 

Example usage:
```
@max_inner_dims(16,32)
for(int ib=0; ib < BLOCK_M; ++ib; @outer) {
  for(int jb=0; jb < BLOCK_N; ++jb; @outer) {
    ...
    for(int it=0; it < 32; ++it; @inner) {
      for(int jt=0; jt < 16; ++jt; @inner) {
      ...
      }
    }
  }
}
```

When the range of the `@inner` loops within an `@outer` block can be determined at compile-time (e.g., because they are constant), this attribute is added automatically.

## Details

If the outermost `@outer`-loop of a for-loop block is decorated with this OKL attribute, the following C++ attributes are added to the resulting backend kernel definitions:

- CUDA/HIP: `__launch_bounds(X*Y*Z)__`
- OpenCL: `__attribute__((reqd_work_group_size(X, Y, Z)))`
- DPC++: `[[sycl::reqd_work_group_size(X,Y,Z)]]`

The number of arguments given should be greater than or equal to the number of `@inner` loops in the block.

In the case where `@max_inner_dims` is specified and the the range of the `@inner` loops can be determined, the former takes precedence&mdash;overriding the kernel launch bounds that would be added automatically.

## Limitations

This attribute only affects the behaviour of the "launched" backends.
Currently CUDA, HIP, OpenCL, and DPC++ are supported: the Metal backend still needs an implementation.

### DIFF
--- a/src/occa/internal/lang/builtins/attributes.hpp
+++ b/src/occa/internal/lang/builtins/attributes.hpp
@@ -9,6 +9,7 @@
 #include <occa/internal/lang/builtins/attributes/implicitArg.hpp>
 #include <occa/internal/lang/builtins/attributes/inner.hpp>
 #include <occa/internal/lang/builtins/attributes/kernel.hpp>
+#include <occa/internal/lang/builtins/attributes/maxInnerDims.hpp>
 #include <occa/internal/lang/builtins/attributes/outer.hpp>
 #include <occa/internal/lang/builtins/attributes/restrict.hpp>
 #include <occa/internal/lang/builtins/attributes/shared.hpp>
--- a/src/occa/internal/lang/builtins/attributes/maxInnerDims.cpp
+++ b/src/occa/internal/lang/builtins/attributes/maxInnerDims.cpp
@@ -14,16 +14,10 @@ namespace occa {
         return name_;
       }
 
-      bool maxInnerDims::forFunction() const {
-        return true;
-      }
-
       bool maxInnerDims::forStatementType(const int sType) const {
-        return (sType 
-          & (statementType::function | statementType::functionDecl)
-          );
+        return (sType & statementType::for_);
       }
-
+      
       bool maxInnerDims::isValid(const attributeToken_t &attr) const {
         if (attr.kwargs.size()) {
           attr.printError("[@max_inner_dims] does not take kwargs");
--- a/src/occa/internal/lang/builtins/attributes/maxInnerDims.cpp
+++ b/src/occa/internal/lang/builtins/attributes/maxInnerDims.cpp
@@ -1,0 +1,62 @@
+#include <occa/internal/lang/expr.hpp>
+#include <occa/internal/lang/parser.hpp>
+#include <occa/internal/lang/statement.hpp>
+#include <occa/internal/lang/variable.hpp>
+#include <occa/internal/lang/builtins/attributes/maxInnerDims.hpp>
+
+namespace occa {
+  namespace lang {
+    namespace attributes {
+      maxInnerDims::maxInnerDims() {}
+
+      const std::string& maxInnerDims::name() const {
+        static const std::string name_{"max_inner_dims"};
+        return name_;
+      }
+
+      bool maxInnerDims::forFunction() const {
+        return true;
+      }
+
+      bool maxInnerDims::forStatementType(const int sType) const {
+        return (sType 
+          & (statementType::function | statementType::functionDecl)
+          );
+      }
+
+      bool maxInnerDims::isValid(const attributeToken_t &attr) const {
+        if (attr.kwargs.size()) {
+          attr.printError("[@max_inner_dims] does not take kwargs");
+          return false;
+        }
+        const auto argCount{attr.args.size()};
+        if (1 > argCount) {
+          attr.printError("[@max_inner_dims] expects at least one argument");
+          return false;
+        }
+        if(3 < argCount) {
+          attr.printError("[@max_inner_dims] takes at most 3 arguments");
+          return false;
+        }
+
+        for(auto&& arg : attr.args) {
+          exprNode *expr = arg.expr;
+          bool error = !(expr && expr->canEvaluate());
+          if(!error){
+            primitive value = expr->evaluate();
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

**Review by noelchalmers — APPROVED** [review]
This works wonderfully. I've played around with it in the cuda lang test (tests/src/internal/lang/modes/cuda.cpp) by changing the multi-kernel source test to:
```cpp
parseAndPrintSource(
    "const int var[10];\n"
    "void foo() {}\n"
    "int bar(int i) {}\n"
    "@kernel void kernel(@restrict int * arg, const int bar) {\n"
    "  @max_inner_dims(256)\n"
    "  for (int o1 = 0; o1 < O1; ++o1; @outer) {\n"
    "    for (int o0 = 0; o0 < O0; ++o0; @outer) {\n"
    "      @shared int shr[3];\n"
    "      @exclusive int excl;\n"
    "      if (true) {\n"
    "        for (int i1 = 10; i1 < (I1 + 4); i1 += 3; @inner) {\n"
    "          for (int i0 = 0; i0 < I0; ++i0; @inner) {\n"
    "            for (;;) {\n"
    "               excl = i0;\n"
    "            }\n"
    "            for (;;) {\n"
    "               excl = i0;\n"
    "            }\n"
    "          }\n"
    "        }\n"
    "      }\n"
    "    }\n"
    "  }\n"
    "  for (int o1 = 0; o1 < O1; ++o1; @outer(0)) {\n"
    "    for (int o0 = 0; o0 < O0; ++o0; @outer(1)) {\n"
    "      @shared int shr[3];\n"
    "      @exclusive int excl;\n"
    "      if (true) {\n"
    "        for (int i1 = 10; i1 < (I1 + 4); i1 += 3; @inner(1)) {\n"
    "          for (int i0 = 0; i0 < I0; ++i0; @inner(0)) {\n"
    "            for (;;) {\n"
    "               excl = i0;\n"
    "            }\n"
    "            for (;;) {\n"
    "               excl = i0;\n"
    "            }\n"
    "          }\n"
    "        }\n"
    "      }\n"
    "    }\n"
    "  }\n"
    "  @max_inner_dims(1024)\n"
    "  for (int ib = 0; ib < entries; ib += 16; @outer) {\n"
    "    for (int it = 0; it < 16; ++it; @inner) {\n"
    "      const int i = ib + it;\n"
    "      if (i < entries) {\n"
    "        ab[i] = a[i] + b[i];\n"
    "      }\n"
    "    }\n"
    "  }\n"
    "}\n"
  );
```
and it works a treat, with the following device code generated:
```cpp
__constant__ int var[10];

__device__ void foo() {}

__device__ int bar(int i) {}

extern "C" __global__ __launch_bounds__(256) void _occa_kernel_0(int * __restrict__ arg,
                                                                 const int bar) {
  {
    int o1 = 0 + blockIdx.y;
    {
      int o0 = 0 + blockIdx.x;
      __shared__ int shr[3];
      int excl;
      if (true) {
        {
          int i1 = 10 + (3 * threadIdx.y);
          {
            int i0 = 0 + threadIdx.x;
            for (; ; ) {
              excl = i0;
            }
            for (; ; ) {
              excl = i0;
            }
          }
        }
      }
    }
  }
}

extern "C" __global__ void _occa_kernel_1(int * __restrict__ arg,
                                          const int bar) {
  {
    int o1 = 0 + blockIdx.x;
    {
      int o0 = 0 + blockIdx.y;
      __shared__ int shr[3];
      int excl;
      if (true) {
        {
          int i1 = 10 + (3 * threadIdx.y);
          {
            int i0 = 0 + threadIdx.x;
            for (; ; ) {
              excl = i0;
            }
            for (; ; ) {
              excl = i0;
            }
          }
        }
      }
    }
  }
}

extern "C" __global__ __launch_bounds__(1024) void _occa_kernel_2(int * __restrict__ arg,
                                                                  const int bar) {
  {
    int ib = 0 + (16 * blockIdx.x);
    {
      int it = 0 + threadIdx.x;
      const int i = ib + it;
      if (i < entries) {
        ab[i] = a[i] + b[i];
      }
    }
  }
}
```
The `addVectors` kernel in this test also has the launch bounds added automatically. 

I'm not certain of all the tests being run and whether they adequately cover this. I'll leave that to @dmed256 to judge. 

A small stylistic comment, the kernel name lines with the launch_bounds are getting pretty long. Should we consider adding line breaks after `__global__` and `__launch_bounds__` just to help with readability?

PR #533, opened 2021-10-22
Package config with Find module wrappers

### LINKED ISSUE #370: Add occa-config.cmake to CMake
Providing `find_package` in config mode following https://cmake.org/cmake/help/latest/command/find_package.html#full-signature-and-config-mode should make it easier for users of occa to integrate it into their own CMake configuration.

AljenU:
Closes #370.

Generate a CMake package file that can be loaded by downstream projects using the find_package(occa) command.

It should even make it relocatable, though that has not been tested yet.


noelchalmers (changes requested):
Works with HIP and OpenCL on the AMD side after small fix

noelchalmers:
Great work on this BTW. This is pretty slick :fire: :fire:

noelchalmers (approved):

PR #539, opened 2021-11-22
[Fortran] Use mpi_f08 module to fix Intel compiler warnings

SFrijters:
## Description

Intel compiler version 19.1.0.20200306 (and above, presumably) complains:

`error #8889: Explicit declaration of the EXTERNAL attribute is required`

This can be solved by using the newer `mpi_f08` module and accompanying newer datatypes.

Open question: if this change is acceptable, should https://github.com/libocca/occa/blob/main/scripts/compiler/fortranCompilerSupportsMPI.f90 also use `mpi_f08`?

Alternative: add `-warn noexternal` to compiler flags to suppress the warning (this was the 'fix' I had in place locally before trying to actually fix it).


kris-rowe:
Looks good. 

> Open question: if this change is acceptable, should https://github.com/libocca/occa/blob/main/scripts/compiler/fortranCompilerSupportsMPI.f90 also use mpi_f08?

Using the mpi_f08 module is safer anyway since it uses explicit interfaces, allowing for type checking. One possible caution against blanket changing this could be if someone is using an older version of OpenMPI or MPICH that doesn't include the mpi_f08 module. 

It would probably be safest to have a second fortranCompilerSupportsMPIF08.f90 file for shellTools.sh to use. That being said, after a quick grep through the code it doesn't appear that fCompilerSupportsMPI in shellTools.sh is called anywhere currently&mdash; so maybe that can go on the backburner for now.

SFrijters:
That pretty much follows my thought process exactly.
Do you know what the intended use of `fCompilerSupportsMPI` is?

kris-rowe:
Most of the functions in shellTools.sh are used by the Makefile build system.

PR #542, opened 2021-11-23
[CI] Add oneAPI compilers (Intel/LLVM) to testing matrix.

kris-rowe:
## Description

- Adds the oneAPI compilers to the testing matrix used by GitHub. 
- The context added runs on the latest ubuntu using the CMake build. 
- Compilers are retrieved using apt.
- Enables CPU testing of the OpenMP, OpenCL, and DPC++ modes
- Adds the OpenCL and DPC++ modes to codecov


SFrijters:
Just an additional note: I've recently started preparing our in-house libraries for IntelLLVM as well and one of the issues I ran into is that CMake doesn't properly detect the compiler before 3.20 (https://cmake.org/cmake/help/latest/release/3.20.html#compilers). I've added a compatibility workaround to our CMake, maybe that could be useful for occa as well if occa needs compiler-specific flags at some point.

PR #555, opened 2022-03-09
OpenMP corrected detection

wjhorne:
## Description
I have been building a binding library between Rust and Occa and ran into an issue when testing OpenMP that this PR fixes. There is probably a better way than this PR to do it. 

Essentially the compiler and subsequent OpenMP flags can be missed easily relative to a Serial device which can create strange behavior. In my case no OpenMP flags would be added by default which resulted in every kernel running in serial. Adding a compiler property to the device creation json wasn't very practical for general use with Rust binding to different c/cpp compilers.


kris-rowe:
Thanks for catching this and for submitting a bug fix!

PR #592, opened 2022-05-31
Fix -Wbitwise-instead-of-logical with new IntelLLVM 2022.1.0

SFrijters:
## Description

Trying to fix the failing CI "Build / [Ubuntu] CMake + Intel/LLVM"

Alternative would be to cast either argument to `int`, but this should have the same effect for bools, right? And it looks slightly more readable.


SFrijters:
Oh, nevermind, I see a different fix is in place already.

PR #606, opened 2022-07-05
Add C API to get CUDA/HIP arch version from device

SFrijters:
## Description

As discussed in the overtime of https://github.com/libocca/occa-taf/blob/main/meeting-notes/2022-06-29.md, this is an example of the kind of information that we would like to be available.
It punches through abstractions rather inelegantly so I'm not expecting this to be accepted as-is and I'll keep it as a draft to be used as input for future development?


SFrijters:
Superseded by #670 .

PR #609, opened 2022-07-25
Feature device synchronization

### LINKED ISSUE #598: Function to synchronize an entire device.
The existing implementation of `device::finish()` only synchronizes the current stream (e.g., calling `cuStreamSynchronize`), making both the function name and documentation somewhat misleading.

Some downstream OCCA applications require a mechanism to wait for all enqueued operations on a device to finish, similar to [`cudaDeviceSynchronize`](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#explicit-synchronization).

The programming models of the other backends (i.e., OpenCL, SYCL) don't have a similar API for device synchronization, however `modeDevice_t` already retains [a vector of streams which have been allocated](https://github.com/libocca/occa/blob/6289743eeb20436ed95d9d9994fbaeb76f393a1a/src/occa/internal/core/device.hpp#L23) so this should not be an issue.

Two potential options to move forward with this are: 
1. Change the implementation of `device::finish()` to match its name and documentation, then add a function to the `stream` class for synchronizing only a particular stream (and possibly a shortcut to synch the current stream).
2. Keep the current implementation of `device::finish()`, but update its documentation and add another function `device::finishAll()` which synchronizes all streams on a device.

kris-rowe:
## Description

Closes #598.

This could be split into 2 separate PRs: the first implementing `stream::finish()` and a subsequent one implementing `device::finishAll()`; however, the need for `stream::finish` is less clear outside of the current context.

### DIFF
--- a/examples/cpp/07_streams/main.cpp
+++ b/examples/cpp/07_streams/main.cpp
@@ -38,10 +38,13 @@ int main(int argc, const char **argv) {
   occa::kernel addVectors = occa::buildKernel("addVectors.okl",
                                  "addVectors");
 
+  // Pass this property to make copies non-blocking on the host.
+  occa::json async_copy({{"async", true}});
+
   // These copies will be submitted to the current
   // stream, which is streamA--the default stream.
-  o_a.copyFrom(a);
-  o_b.copyFrom(b);
+  o_a.copyFrom(a,async_copy);
+  o_b.copyFrom(b,async_copy);
   
   // Waits the copies in streamA to complete
   streamA.finish(); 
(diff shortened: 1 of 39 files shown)


deukhyun-cha:
Currently we have occaFinish()/occa::finish() as c/c++ API. If I didn't miss this doesn't seem include the exposure of stream::finish() and device::finishAll() to that end. I am wondering if/how it would come later. BTW this change would make the implementation (and API) much clear IMO.

kris-rowe:
> Currently we have occaFinish()/occa::finish() as c/c++ API. If I didn't miss this doesn't seem include the exposure of stream::finish() and device::finishAll() to that end. I am wondering if/how it would come later. 

Good catch. I was thinking to add this later, however I might as well do it now before it is merged so that it isn't forgotten.

PR #611, opened 2022-08-01
[CMake/Make/bin] Remove vestigial MPI support

SFrijters:
## Description

PR https://github.com/libocca/occa/pull/608 removed MPI support, after this compile_cpp_mpi_example_with_modes was
unused so it can be removed as well.


kris-rowe:
Good catch!

PR #613, opened 2022-08-10
[doc] fix for OKL inner-loop restriction example

kvoronin-intel:
Minor fix for an example for the inner-loop restriction in OKL (inner loops counts should be equal?)


kris-rowe (approved):
Good catch!

PR #616, opened 2022-08-16
CMake find OpenCL

kris-rowe:
## Description

Updates CMake logic for finding OpenCL headers and library: it should no longer be necessary to define the environment variable `OpenCL_ROOT` on most platforms which have CUDA, HIP, or SYCL installed.


SFrijters:
Nice! I was hacking around to get this to work without touching the existing infrastructure, but this seems to work cleanly for me.

PR #626, opened 2022-10-12
[OpenMP] Fix cancelling kernel hash

noelchalmers:
## Description

Fixes a bug where both the openmp device::hash() and device::kernelHash were identical to the serial device hashes, with an additional ^ occa::hash("openmp"). This extra hash term would cancel out later (since they're identical) in device::setupKernelInfo, leading to the serial and openmp kernels having the same hash in the cache.


kris-rowe (approved):
Good catch.

PR #649, opened 2023-02-23
Malachi/zero word issues

### LINKED ISSUE #648: Zero-word allocations do not work with slicing, `copyTo`, `copyFrom`
```cpp
auto o_nothing = device.malloc<double>(0); // okay, doesn't throw any errors
auto o_slice = o_nothing + 0; // hits assertInitialized
std::vector<double> nothing();
o_nothing.copyFrom(nothing.data(), 0); // hits assertInitialized
```

It's a bit annoying to need to put checks around these "no-op" operations.

MalachiTimothyPhillips:
May close #648 

I don't know how to kick off the unit tests, but I was able to confirm in an example that this seems to address the issue I observed.

### DIFF
--- a/tests/src/core/memory.cpp
+++ b/tests/src/core/memory.cpp
@@ -43,6 +43,19 @@ void testMalloc() {
   mem = occa::malloc(bytes, hostPtr, {{"use_host_pointer", false}});
   ASSERT_EQ(mem.ptr<int>()[0], value);
   ASSERT_NEQ(mem.ptr<int>(), hostPtr);
+
+  // test zero byte malloc
+  const occa::udim_t noBytes = 0 * sizeof(int);
+  hostPtr = NULL;
+
+  occa::memory noMem = occa::malloc(noBytes);
+  ASSERT_EQ(noMem.ptr(), (void*) NULL);
+
+  noMem = occa::malloc(noBytes, hostPtr);
+  ASSERT_EQ(noMem.ptr<int>(), hostPtr);
+
+  noMem = occa::malloc(noBytes, hostPtr, {{"use_host_pointer", true}});
+  ASSERT_EQ(noMem.ptr<int>(), hostPtr);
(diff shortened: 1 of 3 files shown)


tests/src/core/memory.cpp
```diff
@@ -43,6 +43,19 @@ void testMalloc() {
   mem = occa::malloc(bytes, hostPtr, {{"use_host_pointer", false}});
   ASSERT_EQ(mem.ptr<int>()[0], value);
   ASSERT_NEQ(mem.ptr<int>(), hostPtr);
+
+  // test zero byte malloc
+  const occa::udim_t noBytes = 0 * sizeof(int);
+  hostPtr = NULL;
+
+  occa::memory noMem = occa::malloc(noBytes);
+  ASSERT_EQ(noMem.ptr(), (void*) NULL);
```
kris-rowe, line 52: This test is failing on a few platforms. I am a bit leery of using `ASSERT_EQ` for pointers here, since it casts to `size_t`. We should probably have a macro for comparing pointer equality using intptr_t or ptrdiff_t.

kris-rowe:
Another complication to this is that OpenCL doesn't allow for the allocation of zero sized buffers and will return an error flag.

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_creating_buffer_objects

MalachiTimothyPhillips:
> Another complication to this is that OpenCL doesn't allow for the allocation of zero sized buffers and will return an error flag.
> 
> https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_creating_buffer_objects

I think it may also be the same case for CUDA. Not sure what the path forward is on this -- punt on #648?

kris-rowe:
Closing this since the solution isn't viable for some backends and an alternative is available in #649.

Issue #657, opened 2023-03-22
Relaxing thread-local constraint for occa device and make it sharable between threads


deukhyun-cha:
Despite that the thread-local constraint currently enforced in OCCA introduces cleaner implementation by eliminating the complexity of making a thread-safe application IMO, it can reduce efficiency of implementing multi-threaded application by requiring to create/manage a device instance in each thread where each thread has kernels to be offloaded. Allowing the threads to share OCCA device would enable more options in stream managements for instance as well as faster development time without the need of creating/managing multiple device instances.

One example from our development would be parallelizing MPI communication with compute kernels. It is common possibility in many computational applications to compute halo regions of the computational sub-domain assigned on each MPI process and exchange the results between the processes while performing the computation of inside halo region (where the computation is self-contained). This requires secondary thread to invoke MPI communication calls as well as the message data packing/unpacking kernels. When the entire computation is offloaded the packing/unpacking kernels should be as well for a performance. It also makes sense those kernels to be invoked in the secondary thread as it depends on the completion of the MPI communication routines. In addition, we want to run those data packing/unpacking kernels on a designed stream so they can run parallel to the compute kernels. A simple illustration of this scenario could be like as below.

![image](https://user-images.githubusercontent.com/55857168/227011188-9ed23d62-5a67-4a2a-8c8d-798ec04883bd.png)

We have a successful demonstration of the scenario with some minimal modification of OCCA code and thought it could be beneficial to have. I think we can even hide this feature behind the build configuration so it can be turned on only when it is required by a user to be utilized responsibly.

PR #672, opened 2023-03-28
[#657] Relax thread-local constraint for the device shrable by multiple threads

deukhyun-cha:
This is a change related to an issue #657. This can be a minimal change possibly being made to enable sharable OCCA device by multiple threads. With this change two threads invoking kernel builds/runs simultaneously could be demonstrated in our application. Unlocking broader support for multi-threading might require more changes, but it would probably be better to have that happens when an actual case presented to ensure necessary changes and testing IMO. As suggested in the issue, this feature is currently hidden behind a cmake build option (turned off by default), so it is required to be turned on by a user to enable it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,14 @@ if(ENABLE_FORTRAN)
   enable_language(Fortran)
 endif()
 
+option(ENABLE_SHARABLE_DEVICE  "Enable sharable device by multiple threads"  OFF)
+if (ENABLE_SHARABLE_DEVICE)
+  add_compile_definitions(OCCA_THREAD_SHARABLE_ENABLED=1)
+  message("-- OCCA sharable by multi-threads     : Enabled")
+else()
+  add_compile_definitions(OCCA_THREAD_SHARABLE_ENABLED=0)
+endif()
+
 set(OCCA_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set(OCCA_BUILD_DIR  ${CMAKE_BINARY_DIR})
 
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -41,6 +41,14 @@ if(ENABLE_FORTRAN)
   enable_language(Fortran)
 endif()
 
+option(ENABLE_SHARABLE_DEVICE  "Enable sharable device by multiple threads"  OFF)
+if (ENABLE_SHARABLE_DEVICE)
+  add_compile_definitions(ENABLE_THREAD_SHARABLE_OCCA=1)
+  message("-- OCCA sharable by multi-threads     : Eabled")
```
SFrijters, line 47: Eabled -> Enabled

deukhyun-cha: Thanks @SFrijters, corrected typo.

noelchalmers:
Tested on some OCCA projects I have. LGTM

deukhyun-cha:
Note: As this feature is hidden by a CMake flag, it will not run its test in CI/CD pipeline for now. We can turn this on in the pipeline after we have more extensive practical run on our side of the applications when this is merged. Just making a note here.

amikstcyr:
Hello - any decision concerning this PR?

kris-rowe (changes requested):
For consistency, the build option should be added to the Makefile like in the CMake build.

deukhyun-cha:
Updated the name of compiler define used for this feature to be more consistent with others, and added it to Makefile as well.

kris-rowe (approved):

Makefile
```diff
@@ -96,6 +96,7 @@ MAKE_COMPILED_DEFINES := $(shell cat "$(OCCA_DIR)/scripts/build/compiledDefinesT
                                       s,@@OCCA_OPENCL_ENABLED@@,$(OCCA_OPENCL_ENABLED),g;\
                                       s,@@OCCA_METAL_ENABLED@@,$(OCCA_METAL_ENABLED),g;\
 				      												s,@@OCCA_DPCPP_ENABLED@@,$(OCCA_DPCPP_ENABLED),g;\
```
kris-rowe, line 98: Probably an extra double-quote here. Wondering if this is causing the CI tests to fail.

kris-rowe (changes requested):

Makefile
deukhyun-cha: Thanks @kris-rowe. Updated.

kris-rowe (approved):

PR #675, opened 2023-03-29
Add kernel/include_occa and kernel/link_occa properties

deukhyun-cha:
Add kernel/include_occa and kernel/link_occa properties to control including and linking occa into kernels. This PR is a replica of the original PR #603 by @SFrijters with very minor cleanup (removed changes already merged in, and squash the original commits), just in case if there would be any further changes required.

Original description by @SFrijters in the PR was:

**Recently I have been debugging a segmentation fault in our code that seems to have been caused by transitive dependencies of occa. During this process I noticed that all kernels get linked to libocca and occa headers are #included at the top of the kernel code. However, none of our kernels require anything from inside occa, so I think it would be nice to be able to control this. Not linking in occa speeds things up a little bit and potentially prevents unexpected behaviour.

This PR keeps the default behaviour as-is, but makes linking to occa, and #include(i)ng the occa headers and namespace opt-out per kernel. This seems to work well in our code at this moment. Is this something that seems generally useful enough to include?

Open questions: should we do some bikeshedding on the flag name and where should this new option be documented? And I'm not quite sure if the markdown files in docs should just be edited directly or if there is some kind of automation for that?
Remark: I have not been able to test with HIP, but the change looks straightforward enough (famous last words).

I also noticed some inconsistencies in how the kernelHash is calculated, based on what goes into the compiler invocation so this is a separate fix commit. It could be taken out of this PR if requested.**

Please refer the original PR #603 for other discussions has been made.

### DIFF
--- a/examples/cpp/18_nonblocking_streams/main.cpp
+++ b/examples/cpp/18_nonblocking_streams/main.cpp
@@ -27,6 +27,7 @@ int main(int argc, const char **argv) {
   occa::json kernelProps({
     {"defines/block", block},
     {"defines/group", group},
+    {"serial/include_std", true},
   });
   occa::kernel powerOfPi = occa::buildKernel("powerOfPi.okl",
                                              "powerOfPi",
--- a/src/loops/typelessForLoop.cpp
+++ b/src/loops/typelessForLoop.cpp
@@ -34,6 +34,8 @@ namespace occa {
       loopScope.device = device;
     }
 
+    loopScope.props["kernel/include_occa"] = true;
+
     const int outerIterationCount = (int) outerIterations.size();
     const int innerIterationCount = (int) innerIterations.size();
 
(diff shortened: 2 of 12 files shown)


noelchalmers:
I've tested this in some OCCA projects. This certainly will be a breaking change for many, as we'll need to be explicit about including headers like <cmath>, but I think that workable if we're clear about it in release notes.

amikstcyr:
Any decision was made concerning this PR?

kris-rowe (changes requested):
A minor revision: the kernel hashing and compilation in the OpenCL and SYCL backends should be updated like the CUDA and HIP backends.

deukhyun-cha:
Thanks Kris. now the opencl and sycl backends are updated.

kris-rowe (approved):

PR #678, opened 2023-04-22
[Core] Zero size memory quick returns

noelchalmers:
## Description

Let some memory methods return quickly instead of error when they are zero size allocations.

Addresses #648, but in a different manner than PR #649. 

@MalachiTimothyPhillips can you weight in if this would solve your issue?

### DIFF
--- a/src/core/memory.cpp
+++ b/src/core/memory.cpp
@@ -164,7 +164,7 @@ namespace occa {
 
   occa::memory memory::slice(const dim_t offset,
                              const dim_t count) const {
-    assertInitialized();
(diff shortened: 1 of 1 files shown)


kris-rowe:
Since [these changes were included in nekRS](https://github.com/Nek5000/nekRS/commit/0129ca9330583df94c665d068e04d3375eb117d3#diff-cce3d86f807b94ce382ce8eb0672e531480ab6b585e016df1f0cd5297adc3d6c) after this PR was opened, I am going to assume that this solution works for them.

PR #698, opened 2023-07-27
Use gcnArchName instead of gcnArch. 

cjatin:
gcnArch has been deprecated for quite some time now.

gcnArch being an int type can not handle the names like gfx90a.

ROCm 6.0 will remove that particular entry all together.

### DIFF
--- a/src/occa/internal/modes/hip/polyfill.hpp
+++ b/src/occa/internal/modes/hip/polyfill.hpp
@@ -37,7 +37,6 @@ namespace occa {
     char *name;
     size_t totalGlobalMem;
     int maxThreadsPerBlock;
-    int gcnArch;
     char gcnArchName[256];
     int major;
     int minor;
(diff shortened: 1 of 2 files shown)


noelchalmers:
I'm confused. Aren't we already using gcnArchName, and it's guarded by a check that ROCm is >= v3.6 ? OCCA works on gfx90a already AFAIK

mangupta:
@noelchalmers : Yes by default OCCA is using gcnArchName, but OCCA still has gcnArch in the code. As mentioned in https://rocm.docs.amd.com/en/latest/release.html#upcoming-changes-in-future-release, gcnArch is being removed from the device properties in an upcoming major release. So OCCA code needs to be cleaned up. Specifically the if check on gcnArch.

noelchalmers:
Ah, I was being thick and didn't notice the unguarded use of gcnArch at line 60. Sorry about the noise.

noelchalmers (approved):

dmcdougall (approved):

PR #700, opened 2023-08-17
Include original file name in the cached sources

mkbosmans:
This aids debugging, e.g. in stack traces or when setting breakpoints.


noelchalmers:
Big fan of this. Thanks for contributing!

PR #705, opened 2023-08-24
OCCA enum and union support 

kian-ohara:
## Added support for (typedef) enums and (typedef) unions.


mkbosmans:
We have tested this on our side and it correctly handles enums and unions now in kernel code.

mkbosmans:
Would this be still in time to go in the 1.6 release?

PR #718, opened 2023-11-14
Allow user to define the maximum number of kernel arguments

deukhyun-cha:
Allow user to define the maximum number of kernel arguments as a CMake parameter and auto-generate the necessary codes if needed. This is to provide flexibility for any larger number of kernel argument required by an application and to optimize the compile time. When the default value predefined is used it compile with generated codes already there without re-generating them.


deukhyun-cha:
There is a couple of remarks;

- Using Makefile requires defining the variable and running the python script manually if needed. Applying the same default value would be useful.
- Minimum required version of python to run the python script for code generation was obtained with a limited test set (ex. 3.5.1 doesn't work, 3.7.2 works). This could probably be improved.

kris-rowe (approved):

PR #725, opened 2023-12-01
Improve user-definable max number of kernel arguments and automatic code generation.

deukhyun-cha:
This update fixes an issue that each cmake configuration triggers automatic code generation without any changes in cmake options when a user-defined maximum number of kernel arguments is specified. Also added a cmake option that user can specify to generate codes for JIT kernel launch only if one knows that's what is going to be used, so it wouldn't suffer from longer compile time when a large value for the maximum number of kernel arguments is required.


deukhyun-cha:
For the build with Makefile, it copies the default *codegen files in the script directory back to proper places in source tree, so it can compile (while the cmake will copy those into build directory and include them from there).

PR #736, opened 2024-01-24
A couple of Examples/concurrency

deukhyun-cha:
Examples for 1) OpenMP streams to generalize the "stream" concept for all modes 2) Internal thread safe block for occa properties.


deukhyun-cha:
- Created as a draft for any discussion before (or to decide) a merge.
- The original examples/concurrency branch seems behind for some commits from the development and needs to catch up.

Issue #512, opened 2021-09-09
Add support a stream to synchronize for a specific event


deukhyun:
Currently only stream synchronization mechanisms available are ones equivalent to cuStreamSynchronize and cuEventSynchronize. It would be nice to have the one equivalent to cuStreamWaitEvent which can be useful to make multi-stream application more efficient.

kris-rowe:
- CUDA: [cuStreamWaitEvent](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__STREAM.html#group__CUDA__STREAM_1g6a898b652dfc6aa1d5c8d97062618b2f)
- HIP: [hipStreamWaitEvent](https://docs.amd.com/bundle/HIP-API-Guide-v5.4.1/page/a00183.html#gacdd84c8f8ef1539c96c57c1d5bcae633)
- SYCL: [sycl::queue::ext_oneapi_submit_barrier](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_enqueue_barrier.asciidoc)
- OpenCL: [clEnqueueBarrierWithWaitList](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueBarrierWithWaitList)
- Metal: ?

noelchalmers:
For Metal, these might be useful:
[MTLEvent](https://developer.apple.com/documentation/metal/mtlevent)
[encodeSignalEvent](https://developer.apple.com/documentation/metal/mtlcommandbuffer/2966542-encodesignalevent)
[encodeWaitForEvent](https://developer.apple.com/documentation/metal/mtlcommandbuffer/2966543-encodewaitforevent)

See also [Synchronizing Events Within a Single Device](https://developer.apple.com/documentation/metal/resource_synchronization/synchronizing_events_within_a_single_device). I'm not volunteering to write this part, though, since I have no way to test it.

deukhyun-cha:
I appreciate the comments. I don't have environment available for testing Metal as well, but perhaps it might be a good first step to get some build/test instruction updated from anyone running OCCA on the environment if possible (unless it's already available and I missed it). That would probably make things easier for anyone to jump-in and volunteer - including myself :)

PR #746, opened 2024-03-18
Add a function for stream to wait for a tag

deukhyun-cha:
This is to implement a feature described in Add support a stream to synchronize for a specific event #512.

### DIFF
--- a/examples/cpp/19_stream_tags/.gitignore
+++ b/examples/cpp/19_stream_tags/.gitignore
@@ -1,0 +1,3 @@
+main
+main.o
+main_c
--- a/examples/cpp/19_stream_tags/CMakeLists.txt
+++ b/examples/cpp/19_stream_tags/CMakeLists.txt
@@ -1,0 +1,4 @@
+compile_cpp_example_with_modes(stream_tags main.cpp)
+
+add_custom_target(cpp_example_stream_tags_okl ALL COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/powerOfPi2.okl powerOfPi2.okl)
+add_dependencies(examples_cpp_stream_tags cpp_example_stream_tags_okl)
--- a/examples/cpp/19_stream_tags/Makefile
+++ b/examples/cpp/19_stream_tags/Makefile
@@ -1,0 +1,27 @@
+
+PROJ_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+
+ifndef OCCA_DIR
+  include $(PROJ_DIR)/../../../scripts/build/Makefile
+else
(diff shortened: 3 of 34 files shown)


deukhyun-cha:
I would appreciate some help on testing especially for dpcpp and metal backends if possible.

deukhyun-cha:
> I would appreciate some help on testing especially for dpcpp and metal backends if possible.

I was able to run the example added here for serial, cuda, dpcpp, and openmp backends.

deukhyun-cha:
I was going through this change and feeling the interface should be done through the device like other stream related functions. If so, should it also work with current stream on the device (instead of taking the stream as an argument)? I would appreciate advice.

kris-rowe:
> I was going through this change and feeling the interface should be done through the device like other stream related functions. If so, should it also work with current stream on the device (instead of taking the stream as an argument)? I would appreciate advice.

Since it is complete, I would leave the existing implementation in the PR as is.

There is already a [`waitFor` function in `device`](https://github.com/libocca/occa/blob/3cba0841b2b87678da53e0b311cb7e162d781181/include/occa/core/device.hpp#L482) that waits on a given `streamTag`. Since `streamTags` are associated with the `stream` which was active when they were created, the existing implementation of `device::waitFor` is problematic in the case where the active stream has been changed by the user. To avoid this ambiguity, `device::waitFor` should be reimplemented later using the function added in this PR to synchronize the current `stream`. A separate API for waiting on a specific `streamTag` could be added to that class afterwards, if needed.

deukhyun-cha:
> > I was going through this change and feeling the interface should be done through the device like other stream related functions. If so, should it also work with current stream on the device (instead of taking the stream as an argument)? I would appreciate advice.
> 
> Since it is complete, I would leave the existing implementation in the PR as is.
> 
> There is already a [`waitFor` function in `device`](https://github.com/libocca/occa/blob/3cba0841b2b87678da53e0b311cb7e162d781181/include/occa/core/device.hpp#L482) that waits on a given `streamTag`. Since `streamTags` are associated with the `stream` which was active when they were created, the existing implementation of `device::waitFor` is problematic in the case where the active stream has been changed by the user. To avoid this ambiguity, `device::waitFor` should be reimplemented later using the function added in this PR to synchronize the current `stream`. A separate API for waiting on a specific `streamTag` could be added to that class afterwards, if needed.

Thanks @kris-rowe, appreciate your comment.

deukhyun-cha:
I was able to build this for Metal backend and run stream examples (after updating for some fixes) on a Mac. The branch is update.

deukhyun-cha:
@kris-rowe just wanted to check if you have any general comments for me to consider on this PR before the detailed review?

kris-rowe (approved):

PR #749, opened 2024-03-29
Intergration of new transpiler with C++ full support

vyast-softserveinc:
## Description

This pull request is aimed for integration occa-transpiler library for providing full C++ support under the OCCA

Added:
* option for switching between old & new transpiler `transpiler-version`
* minor code improvments


src/occa/internal/core/launchedDevice.cpp
```diff
@@ -1,55 +1,132 @@
...
+    }
+
+    auto defines = transpiler::buildDefines(kernelProps);
+    auto includes = transpiler::buildIncludes(kernelProps);
+
+    std::string fullFilePath = io::expandFilename(filename);
+    std::ifstream sourceFile(fullFilePath);
+    std::string sourceCode{std::istreambuf_iterator<char>(sourceFile), {}};
+    oklt::UserInput input {
+        .backend = oklt::TargetBackend::CUDA,
```
vyast-softserveinc, line 52: use targetIter.second

kris-rowe:
@vyast-softserveinc Can you update this PR to target the `development` branch?

kris-rowe:
> @kris-rowe is it possible to setup the llvm 17 for the images? This update depends on the clang 17. The reference for setup from the occa-transpiler is the following: https://github.com/libocca/occa-transpiler/tree/devel?tab=readme-ov-file#setup-llvm

We are doing something similar in the GitHub actions to test the oneAPI compilers ([here](https://github.com/libocca/occa/blob/3cba0841b2b87678da53e0b311cb7e162d781181/.github/workflows/build.yml#L87)). Does it make sense to test the transpiler for all of the configurations we currently do or only a subset?

vyast-softserveinc:
By the last commits I made optional building . So I think it would be enough to make it for single configuration.

vyast-softserveinc:
This PR is out of date. new implementation with the https://github.com/libocca/occa/pull/756 is provided instead

PR #766, opened 2024-09-09
Fixes for metal backend

thilinarmtb:
## Description

Trying to get the Metal backend tests pass on my M2.
The following tests fail as of now:

- [ ]  5 - bin-c-kernel (SEGFAULT)
- [ ] 10 - bin-core-kernel (SEGFAULT)
- [ ] 16 - bin-functional-array (Subprocess aborted)
- [ ] 17 - bin-functional-range (Subprocess aborted)
- [ ] 55 - bin-loops-forLoop (SEGFAULT)

### DIFF
--- a/cmake/FindMETAL.cmake
+++ b/cmake/FindMETAL.cmake
@@ -3,24 +3,24 @@
 # This Find module is also distributed alongside the occa package config file!
 ###############################################################################
 
+find_library(FOUNDATION_LIBRARY Foundation)
 find_library(METAL_LIBRARY Metal)
-find_library(CORE_SERVICES CoreServices)
-find_library(APP_KIT AppKit)
+find_library(METAL_KIT_LIBRARY MetalKit)
 
 include(FindPackageHandleStandardArgs)
 find_package_handle_standard_args(
     METAL
     REQUIRED_VARS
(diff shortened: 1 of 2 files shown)


thilinarmtb:
Seems like these test failures happen even with the Serial backend on M2.

thilinarmtb:
I am investigating the failure of `5-bin-c-kernel`. Seems like if the generated
C source for `argKernel` (`tests/files/argKernel.okl`) has the following
function signature, the execution fails on M2 with the serial backend.
```C++
extern "C" void argKernel(void *nullPtr, int *mem, char &i8,
                            unsigned char &u8, short &i16, unsigned short &u16,
                            int &i32, unsigned int &u32, long int &i64,
                            unsigned long int &u64, float &f, double &d,
                            mystruct &xy, const char *str);
```

Instead, if it has the following (without references), I could get it to work with
`dlopen()` + `dlsym()`.
```C++
extern "C" void argKernel(void *nullPtr, int *mem, char i8, unsigned char u8,
                            short i16, unsigned short u16, int i32,
                            unsigned int u32, long int i64,
                            unsigned long int u64, float f, double d,
                            mystruct xy, const char *str);
```

PR #770, opened 2024-10-30
Fix a bug in parsing occa cli options

thilinarmtb:
## Description

With the current `development` branch, the following doesn't work (it produces serial output):
```sh
./bin/occa translate ~/addVectors.okl --mode cuda
```

With this change, the above work as expected and produces cuda output.

### Todo
- [x] Add a test for the failed case

### DIFF
--- a/src/occa/internal/bin/occa.hpp
+++ b/src/occa/internal/bin/occa.hpp
@@ -6,6 +6,7 @@
 namespace occa {
   namespace bin {
     cli::command buildOccaCommand();
+    bool runTranslate(const json &args);
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/occa/internal/bin/occa.hpp
```diff
@@ -6,6 +6,7 @@
 namespace occa {
   namespace bin {
     cli::command buildOccaCommand();
+    bool runTranslate(const json &args);
```
thilinarmtb, line 9: Had to expose this function so I don't have to redundantly implement it again in tests. Hope that is okay.

kris-rowe (approved):

PR #773, opened 2024-11-18
Update CUDA backend to create streams with priority

deukhyun-cha:
Update CUDA backend to create streams with priority. In Jason property for a stream creation a user can turn on & specify a custom priority level based on a user defined priority range also defined in the property. This way the user can be hidden from the actual CUDA priority range. The priority level and range are defined in integers.

### DIFF
--- a/src/occa/internal/modes/cuda/device.cpp
+++ b/src/occa/internal/modes/cuda/device.cpp
@@ -120,8 +120,24 @@ namespace occa {
       setCudaContext();
 
       if (props.get<bool>("nonblocking", false)) {
-        OCCA_CUDA_ERROR("Device: createStream - NonBlocking",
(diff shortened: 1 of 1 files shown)


kris-rowe:
Adding these as documentation for future development extending this to other backends:

- [hipStreamCreateWithPriority](https://rocm.docs.amd.com/projects/HIP/en/latest/reference/hip_runtime_api/modules/stream_management.html#_CPPv427hipStreamCreateWithPriorityP11hipStream_tji)
- [oneAPI queue priority extension for SYCL](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_queue_priority.asciidoc)
- [OpenCL queue priority extension](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#cl_khr_priority_hints)

PR #777, opened 2024-11-21
Avoid `-fsycl` being added to non-SYCL code

thilinarmtb:
## Description

In the current `development` version, `FindDPCPP.cmake` adds `-fsycl` to the
host (one that is being used to compile OCCA) compiler. This results in compilation
errors if the host compiler doesn't support SYCL. There is no need for the host
compiler to support SYCL to use the OCCA SYCL backend. Users simply has to
set the following variables:

```sh
export OCCA_DPCPP_COMPILER=clang++
export OCCA_DPCPP_COMPILER_FLAGS="-fsycl -fsycl-targets=nvptx64-nvidia-cuda"
```

### DIFF
--- a/cmake/FindDPCPP.cmake
+++ b/cmake/FindDPCPP.cmake
@@ -9,7 +9,7 @@ unset(missingDpcppComponents)
 find_path(
   SYCL_INCLUDE_DIRS
   NAMES
-    sycl.hpp
+    sycl/sycl.hpp
   PATHS
     ENV SYCL_ROOT
     /opt/intel/oneapi/compiler/latest/linux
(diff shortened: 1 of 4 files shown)


cmake/FindDPCPP.cmake
```diff
@@ -29,34 +28,21 @@ find_library(
...
 
 if(DPCPP_FOUND AND NOT TARGET OCCA::depends::DPCPP)
   # Create our wrapper imported target
   # Put it in the OCCA namespace to make it clear that we created it.
   add_library(OCCA::depends::DPCPP INTERFACE IMPORTED)
-  separate_arguments(SYCL_FLAGS UNIX_COMMAND "${SYCL_FLAGS}")
-  target_compile_options(OCCA::depends::DPCPP INTERFACE ${SYCL_FLAGS})
   set_target_properties(OCCA::depends::DPCPP PROPERTIES
-    INTERFACE_INCLUDE_DIRECTORIES "${SYCL_INCLUDE_DIRS}"
+    INTERFACE_INCLUDE_DIRECTORIES "${SYCL_INCLUDE_DIR};${SYCL_INCLUDE_DIR}/sycl"
```
thilinarmtb: Should we add `${SYCL_INCLUDE_DIR}/CL` here as well?

thilinarmtb: Adding `${SYCL_INCLUDE_DIR}/CL` gives an error as it is non-existent.

thilinarmtb: Maybe we should just add `${SYCL_INCLUDE_DIR}` to the interface. 
I will remove the `${SYCL_INCLUDE_DIR}/sycl` from the interface too.

thilinarmtb:
This works on Aurora as is. I will check again on Polaris when it comes back live.

thilinarmtb:
After discussions with @kris-rowe, now cmake will bail out if DPCPP is enabled but the compiler doesn't support DPCPP.

PR #779, opened 2025-02-28
Check if the compiler supports `std::filesystem`

thilinarmtb:
## Description

We need `std::filesystem` support in the CXX compiler in order to compile OCCA.
Seems like some gcc versions even with `-std=c++17` set doesn't support `std::filesystem`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,12 +80,6 @@ if (OCCA_CLANG_BASED_TRANSPILER)
 endif()
 
 include(SetCompilerFlags)
-include(CheckCXXCompilerFlag)
-
-check_cxx_compiler_flag("-fno-strict-aliasing" COMPILER_SUPPORTS_NO_STRICT_ALIASING)
-if(COMPILER_SUPPORTS_NO_STRICT_ALIASING)
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-strict-aliasing")
-endif()
 
 string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPER)
 message("-- System     : ${CMAKE_SYSTEM}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,8 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+include(CheckRequiredCompilerFeatures)
+
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 option(OCCA_ENABLE_OPENMP "Build with OpenMP if available" ON)
--- a/cmake/CheckRequiredCompilerFeatures.cmake
+++ b/cmake/CheckRequiredCompilerFeatures.cmake
@@ -1,0 +1,15 @@
+include(CheckCXXCompilerFlag)
+
+check_cxx_source_compiles("
+#include <filesystem>
+
+namespace fs = std::filesystem;
+
+int main(int argc, char *argv[]) {
+  return 0;
+}
+" HAS_FILE_SYSTEM)
+
+if (NOT HAS_FILE_SYSTEM)
+  message(FATAL_ERROR "CXX compiler doesn't have std::filesystem support !")
+endif()
--- a/cmake/SetCompilerFlags.cmake
+++ b/cmake/SetCompilerFlags.cmake
@@ -87,3 +87,8 @@ if (OCCA_ENABLE_FORTRAN)
   set_optional_fortran_flag(SUPPORTED_WNO_INTEGER_DIVISION_Fortran_FLAGS "-Wno-integer-division")
 
 endif()
+
+check_cxx_compiler_flag("-fno-strict-aliasing" COMPILER_SUPPORTS_NO_STRICT_ALIASING)
+if(COMPILER_SUPPORTS_NO_STRICT_ALIASING)
+  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-strict-aliasing")
+endif()


thilinarmtb:
This PR fails on GitHub during CI since we have `-Werror` which forces all warnings to be
treated as errors (except for the exceptions we later specify) and `gcc >= 13` finds a false
positive in the occa code:
```C++
/home/runner/work/occa/occa/src/c/device.cpp: In function ‘occaJson occaDeviceGetKernelProperties(occaDevice)’:
/home/runner/work/occa/occa/src/c/device.cpp:51:21: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
   51 |   const occa::json &props = occa::c::device(device).kernelProperties();
```

One option is to just remove `-Werror` (we can still see all the warnings). Other option is to
convert the references to pointers (just copying without the references or pointers lead to
test failures and possibly could have performance impacts as well) with something like the
following:

```diff
diff --git a/include/occa/core/device.hpp b/include/occa/core/device.hpp
index b0461f23..4e1e8e5f 100644
--- a/include/occa/core/device.hpp
+++ b/include/occa/core/device.hpp
@@ -281,6 +281,7 @@ namespace occa {
      * @endDoc
      */
     const occa::json& properties() const;
+    const occa::json* propertiesPtr() const;

     /**
      * @startDoc{arch}
@@ -298,12 +299,15 @@ namespace occa {
     const std::string& arch() const;

     const occa::json& kernelProperties() const;
+    const occa::json* kernelPropertiesPtr() const;
     occa::json kernelProperties(const occa::json &additionalProps) const;

     const occa::json& memoryProperties() const;
+    const occa::json* memoryPropertiesPtr() const;
     occa::json memoryProperties(const occa::json &additionalProps) const;

     const occa::json& streamProperties() const;
+    const occa::json* streamPropertiesPtr() const;
     occa::json streamProperties(const occa::json &additionalProps) const;

     /**
diff --git a/src/core/device.cpp b/src/core/device.cpp
index 45fbdd42..f8dbb902 100644
--- a/src/core/device.cpp
+++ b/src/core/device.cpp
@@ -176,6 +176,11 @@ namespace occa {
     return modeDevice->properties;
   }

+  const occa::json* device::propertiesPtr() const {
+    assertInitialized();
+    return &(modeDevice->properties);
+  }
+
   const std::string& device::arch() const {
     static const std::string noArch = "No Arch";
     return (modeDevice
@@ -188,6 +193,11 @@ namespace occa {
     return (const occa::json&) modeDevice->properties["kernel"];
   }

+  const occa::json* device::kernelPropertiesPtr() const {
+    assertInitialized();
+    return (const occa::json*) &(modeDevice->properties["kernel"]);
+  }
+
   occa::json device::kernelProperties(const occa::json &additionalProps) const {
     return (
       kernelProperties()
@@ -200,6 +210,11 @@ namespace occa {
     return (const occa::json&) modeDevice->properties["memory"];
   }

+  const occa::json* device::memoryPropertiesPtr() const {
+    assertInitialized();
+    return (const occa::json*) &(modeDevice->properties["memory"]);
+  }
+
   occa::json device::memoryProperties(const occa::json &additionalProps) const {
     return (
       memoryProperties()
@@ -212,6 +227,11 @@ namespace occa {
     return (const occa::json&) modeDevice->properties["stream"];
   }

+  const occa::json* device::streamPropertiesPtr() const {
+    assertInitialized();
+    return (const occa::json*) &(modeDevice->properties["stream"]);
+  }
+
   occa::json device::streamProperties(const occa::json &additionalProps) const {
     return (
       streamProperties()
diff --git a/src/c/device.cpp b/src/c/device.cpp
index 475f3467..c424e1a8 100644
--- a/src/c/device.cpp
+++ b/src/c/device.cpp
@@ -39,8 +39,8 @@ const char* occaDeviceMode(occaDevice device) {
 }

 occaJson occaDeviceGetProperties(occaDevice device) {
-  const occa::json &props = occa::c::device(device).properties();
-  return occa::c::newOccaType(props, false);
+  const occa::json *props = occa::c::device(device).propertiesPtr();
+  return occa::c::newOccaType(*props, false);
 }

 const char* occaDeviceArch(occaDevice device) {
@@ -48,18 +48,18 @@ const char* occaDeviceArch(occaDevice device) {
 }

 occaJson occaDeviceGetKernelProperties(occaDevice device) {
-  const occa::json &props = occa::c::device(device).kernelProperties();
-  return occa::c::newOccaType(props, false);
+  const occa::json *props = occa::c::device(device).kernelPropertiesPtr();
+  return occa::c::newOccaType(*props, false);
 }

 occaJson occaDeviceGetMemoryProperties(occaDevice device) {
-  const occa::json &props = occa::c::device(device).memoryProperties();
-  return occa::c::newOccaType(props, false);
+  const occa::json *props = occa::c::device(device).memoryPropertiesPtr();
+  return occa::c::newOccaType(*props, false);
 }

 occaJson occaDeviceGetStreamProperties(occaDevice device) {
-  const occa::json &props = occa::c::device(device).streamProperties();
-  return occa::c::newOccaType(props, false);
+  const occa::json *props = occa::c::device(device).streamPropertiesPtr();
+  return occa::c::newOccaType(*props, false);
 }

 occaUDim_t occaDeviceMemorySize(occaDevice device) {
```

kris-rowe:
It seems like pointers are the way to go, especially since the [underlying C struct is storing the reference as a pointer](https://github.com/libocca/occa/blob/65efd39c1fd1fe1de0317f4ed5b176bd142303dc/include/occa/c/types.h#L39) anyway.
We should address this in a separate PR as a way of documenting things.

thilinarmtb:
> It seems like pointers are the way to go, especially since the [underlying C struct is storing the reference as a pointer](https://github.com/libocca/occa/blob/65efd39c1fd1fe1de0317f4ed5b176bd142303dc/include/occa/c/types.h#L39) anyway. We should address this in a separate PR as a way of documenting things.

Fixed in #783. Seems like the usage of pointers wasn't necessary as we could simply avoid using the temporary variable.

PR #782, opened 2025-06-02
Remove deprecated GitHub runner

kris-rowe:


barracuda156:
@kris-rowe No direct impact for me, but just for the info, why?

kris-rowe:
Given the bodies we have to support occa at the moment and that most of the applications within our scope run on large Linux systems, we dropped testing on the macOS runners so that this didn't detract from other work.

PR #784, opened 2025-06-09
Validate CITATION.cff

sejalgupta11:
## Description
Removed "publisher" field to meet cffconvert guidelines, now gets clean cffconvert report


thilinarmtb:
@sejalgupta11 : Maybe we can put the 2014 paper (which is what the `CITATION.cff` is about right
now) under `preffered-citation` and then add the core library contributors under `authors` as in
[here](https://github.com/CEED/libCEED/blob/main/CITATION.cff#L59C1-L59C19)?

We can then fill up the [other sections](https://github.com/CEED/libCEED/blob/main/CITATION.cff#L2C1-L5C23) with the data related to the OCCA software releases in GitHub.

thilinarmtb:
@sejalgupta11 : Should we close this in favor of #785 ?

PR #793, opened 2025-06-25
Update cmake build system

thilinarmtb:
## Description

Use `add_subdirectory()` instead of `GLOB_RECURSE` to add source files to the build system.
This will enable exposing part of internal code (`cli`, `utils`, etc.) to OCCA related projects
if necessary.


thilinarmtb:
I did some timing comparing this branch and the development. There is no measurable slowdown
due to the changes.

|  | configure | build|
|--|---|--|
|This PR|11.8s|46.6s|
|development|11.2s|46.4s|
______________________

PR #794, opened 2025-07-01
Add env var to override OCCA HIP compile arch

dmcdougall:
## Description

This commit teaches OCCA to understand the `OCCA_HIP_OVERRIDE_COMPILE_ARCH` environment variable.  If set, OCCA will ignore the arch set in the kernel properties and try to compile for the architecture provided in the environment variable.

This allows the user to have more control over what targets are JITted for. Usecases are for JITting for generic targets (see [here](https://github.com/llvm/llvm-project/commit/f93aa5157a3317b24cff660ac972814ee9ed4dbc)) and for the (currently beta) amdgcnspirv target (see [here](https://rocm.docs.amd.com/projects/llvm-project/en/latest/conceptual/spirv.html#rocm-support-for-spir-v-beta)).

Unfortunately, the amdgcnspirv target name doesn't start with "gfx", so I added a branch that checks for this case and treats it accordingly.

Comments, criticism, tomatoes, and hecklers are not only welcome but heavily encouraged.

### DIFF
--- a/src/occa/internal/modes/hip/device.cpp
+++ b/src/occa/internal/modes/hip/device.cpp
@@ -220,10 +220,15 @@ namespace occa {
 
         std::string archString = kernelProps.get<std::string>("arch", arch);
 
+        // Override the arch if the users requests it
+        if (env::var("OCCA_HIP_OVERRIDE_COMPILE_ARCH").size()) {
+          archString = env::var("OCCA_HIP_OVERRIDE_COMPILE_ARCH");
+        }
+
         std::string archFlag;
         if (startsWith(archString, "sm_")) {
           archFlag = " -arch=" + archString;
-        } else if (startsWith(archString, "gfx")) {
+        } else if (startsWith(archString, "gfx") || startsWith(archString, "amd")) {
           archFlag = " --offload-arch=" + archString;
         } else {
           OCCA_FORCE_ERROR("Unknown HIP arch");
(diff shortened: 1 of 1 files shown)


thilinarmtb:
@dmcdougall : Seems like both CUDA and HIP both set the architecture compiler option. Wondering
if it is a good idea to rename the variable to `OCCA_OVERRIDE_COMPILE_ARCH` and let the backends which
set this compiler option query the env var (or better yet we update the kernelProperties based on the
env var and pass it around).

kris-rowe:
@dmcdougall Could you update this to target the development branch?

dmcdougall:
> Could you update this to target the development branch?

Rebased on top of `development` and now targets `development`.

dmcdougall:
> Seems like both CUDA and HIP both set the architecture compiler option.

Honestly, I think any backend should be able to do this.

> Wondering if it is a good idea to rename the variable to OCCA_OVERRIDE_COMPILE_ARCH and let the backends which
set this compiler option query the env var

I totally agree.  I had initially wanted to do it this way, but didn't see a way for occa to consume the variable in a place that wasn't backend-specific.

> (or better yet we update the kernelProperties based on the env var and pass it around).

Ah yes, maybe this is they way I should have done it to begin with.

dmcdougall:
Before I continue with implementing @thilinarmtb's feedback, is the general principle behind this idea acceptable?

thilinarmtb:
> Before I continue with implementing @thilinarmtb's feedback, is the general principle behind this idea acceptable?

I think this is a useful environment variable to have. Kris is on vacation so probably we will have to
wait till next week to know his opinion.

dmcdougall:
I decided to just rename it instead, and let the backend be responsible for deciding how to populate the kernel properties.

PR #805, opened 2025-11-05
Split libocca target

thilinarmtb:


thilinarmtb:
All the tests pass on Polaris with the following commands:
```sh
module load gcc-native
module use /soft/modulefiles/
module load spack-pe-base cmake
cmake --workflow --preset gnu-default
```

PR #806, opened 2025-11-05
Remove `OCCA_DPCPP_COMPILER` from `CmakePresets.json`

thilinarmtb:
### Todo
- [x] Test on Polaris


thilinarmtb:
Tested on Polaris using the following commands and it works without any issues.
```sh
module use /soft/modulefiles
module load oneapi/upstream
cmake --workflow --preset clang-sycl-nvidia
```

